PR #3, opened 2021-06-22
Fixes from PR#1

ssbarnea:


setup.cfg
```diff
@@ -76,7 +71,7 @@ test =
 where = src
 
 [options.package_data]
-ansiblelint = py.typed
+ansiblecompat = py.typed
```
webknjaz, line 74: unrelated question, but why the importable name is "ansiblecompat" and not "ansible_compat"?

tox.ini
```diff
@@ -11,20 +11,18 @@ usedevelop = false
 
 [testenv]
 description =
-  Run the tests under {basepython} and
```
webknjaz, line 14: Why is "and" unnecessary? isn't it merged with the following lines?

tox.ini
```diff
@@ -100,10 +101,10 @@ commands =
   os.path.isdir(dist_dir) or sys.exit(0); \
   print("Removing \{!s\} contents...".format(dist_dir), file=sys.stderr); \
   shutil.rmtree(dist_dir)'
-  {envpython} -m pep517.build \
-    --source \
-    --binary \
-    --out-dir {toxinidir}/dist/ \
+  {envpython} -m build \
+    --sdist \
+    --wheel \
```
webknjaz, line 106: Drop these. Starting with build v0.5.0, it builds sdist and then builds a wheel from that sdist tarball. But if you specify `--sdist` / `--wheel`, both end up being built from the Git checkout which is not how pip will behave if it hits sdist.
```suggestion
```

setup.cfg
ssbarnea: That is because I followed naming used by `ansiblelint` package. In fact I personally prefer easier to read imports, but at the same time I do not enjoy having some with and some without separator. Do we already have some other packages starting with ansible to compare against?

tox.ini
webknjaz: Note: all the other comments except for this one are non-blockers. Apply this patch and feel free to merge.

setup.cfg
webknjaz: Does https://github.com/ansible/ansible-navigator/tree/main/ansible_navigator count?

webknjaz: Also https://github.com/ansible/ansible-builder/tree/devel/ansible_builder

webknjaz: And finally https://github.com/ansible-community/ansible-bender/tree/master/ansible_bender (off the top of my head)

tox.ini
ssbarnea: Run `tox -va` locally and you will understand. Before making this change it was producing wrong text due to sentence ending on "and" for the simple python envs.

So far that approach was the only one that produced relatively good descriptions:

setup.cfg
webknjaz: Maybe there's others but this should be enough to state that there are projects that follow both practices. But from a Python dev PoV (readability and style guides), I'd prefer an underscore. It's probably best to do this in a separate PR, though.

ssbarnea: I will fix that in follow up, 3 examples was more than enough. I wonder if we do the same for linter in the future.

webknjaz (approved):

Issue #4, opened 2021-06-22
migrate code from ansiblelint.prerun module


ssbarnea:
Move most/all code from `ansiblelint.prerun` into `ansiblecompat` so we can replace molecule dependency on ansiblelint with a dependency on this module, same would apply to ansiblelint.

PR #5, opened 2021-06-24
Import shared code from ansiblelint

ssbarnea:
This change should move all the shared code from ansiblelint and allow removal of molecule dependency on ansible-lint.

Both molecule and ansible-lint with import the code from ansible-compat.

Fixes: #4


src/ansible_compat/config.py
```diff
@@ -0,0 +1,98 @@
...
+)
+
+
+# Used to store collection list paths (with mock paths if needed)
+collection_list: List[str] = []
+
+
+def get_rule_config(rule_id: str) -> Dict[str, Any]:
+    """Get configurations for the rule ``rule_id``."""
+    rule_config = options.rules.get(rule_id, dict())
```
webknjaz, line 32: plz use a literal here

src/ansible_compat/config.py
```diff
@@ -0,0 +1,98 @@
...
+    if not isinstance(rule_config, dict):
+        raise RuntimeError("Invalid rule config for %s: %s" % (rule_id, rule_config))
```
webknjaz, line 34: It's best to name the interpolation placeholders.

src/ansible_compat/config.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def parse_ansible_version(stdout: str) -> Tuple[str, Optional[str]]:
+    """Parse output of 'ansible --version'."""
+    # Ansible can produce extra output before displaying version in debug mode.
+
+    # ansible-core 2.11+: 'ansible [core 2.11.3]'
+    match = re.search(r"^ansible \[(?:core|base) ([^\]]+)\]", stdout, re.MULTILINE)
+    if match:
+        return match.group(1), None
```
webknjaz, line 59: Could you name the group? Having to keep track of numbers adds to the cognitive complexity.

src/ansible_compat/config.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    if not version:
+        proc = subprocess.run(
+            ["ansible", "--version"],
+            universal_newlines=True,
+            check=False,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        if proc.returncode == 0:
```
webknjaz, line 83: Invert this into a guard expression to reduce nesting.

src/ansible_compat/config.py
```diff
@@ -0,0 +1,98 @@
...
+        else:
+            print(
+                "Unable to find a working copy of ansible executable.",
+                proc,
+            )
+            sys.exit(ANSIBLE_MISSING_RC)
+    return Version(version)
+
+
+if ansible_collections_path() in os.environ:
```
webknjaz, line 97: It's best not to have any import-time code execution. You could use `__getattr__()` to make it on-access though.

src/ansible_compat/constants.py
```diff
@@ -0,0 +1,31 @@
...
+)
+
+INVALID_CONFIG_RC = 2
+ANSIBLE_MISSING_RC = 4
+INVALID_PREREQUISITES_RC = 10
+# odict is the base class used to represent data model of Ansible
+# playbooks and tasks.
+odict = dict
+if sys.version_info[:2] < (3, 7):
+    try:
```
webknjaz, line 27: There's a nice suppression idiom for this:
```suggestion
    with contextlib.suppress(ImportError):
```

src/ansible_compat/constants.py
```diff
@@ -0,0 +1,31 @@
...
+    from typing import Literal  # pylint: disable=no-name-in-module
+else:
+    from typing_extensions import Literal
+
+# Minimal version of Ansible we support for runtime
+ANSIBLE_MIN_VERSION = "2.9"
+
+# Based on https://docs.ansible.com/ansible/latest/reference_appendices/config.html
+ANSIBLE_DEFAULT_ROLES_PATH = (
+    "~/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles"
```
webknjaz, line 17: Some software doesn't resolve `~` but resolves `$HOME`. Should we care about it in this string?

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+)
+from ansible_compat.loaders import yaml_from_file
+
+_logger = logging.getLogger(__name__)
+
+
+def check_ansible_presence(exit_on_error: bool = False) -> Tuple[str, str]:
+    """Assures we stop execution if Ansible is missing or outdated.
+
+    Returne found version and an optional exception if something wrong
```
webknjaz, line 37:
```suggestion
    Return found version and an optional exception if something wrong
```

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+import pathlib
+import re
+import subprocess
+import sys
+from functools import lru_cache
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
+
+import packaging
+import tenacity
+from packaging import version
```
webknjaz, line 14: double import of the same pkg

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+        )
+        if result.returncode != 0:
+            return (
+                ver,
+                "FATAL: Unable to retrieve ansible cli version: %s" % result.stdout,
+            )
+
+        ver, error = parse_ansible_version(result.stdout)
+        if error is not None:
+            return "", error
```
webknjaz, line 61: Looks like golang here. We don't do this in Python. If you want to communicate an error, communicate it. Nothing prevents you from attaching the attributes to the exception.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+    )
+    if run.returncode != 0:
+        _logger.error("Command returned %s code:\n%s", run.returncode, run.stdout)
+        sys.exit(INVALID_PREREQUISITES_RC)
+
+
+@tenacity.retry(  # Retry up to 3 times as galaxy server can return errors
```
webknjaz, line 129: Why was this lib chosen? My usual go-to choice is `backoff`. Is it somehow better?

webknjaz: Also, the next version of ansible-core should come with its own retry logic so maybe it's worth making it conditional.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+            _logger.error(run.stdout)
+            raise RuntimeError(run.returncode)
+
+
+def prepare_environment(required_collections: Optional[Dict[str, str]] = None) -> None:
+    """Make dependencies available if needed."""
+    if not options.configured:
+        # Allow method to be used without calling the command line, so we can
+        # reuse it in other tools, like molecule.
+        # pylint: disable=import-outside-toplevel,cyclic-import
```
webknjaz, line 193: if this is unnecessary, don't keep it.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+
+        # initialize_options()
+        pass
+
+    if not options.offline:
+        install_requirements("requirements.yml")
+        for req in pathlib.Path(".").glob("molecule/*/requirements.yml"):
+            install_requirements(str(req))
+
+    if required_collections:
```
webknjaz, line 204: It's better to make this `{}` if `None` at the beginning and save one level of indentation.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+        for name, min_version in required_collections.items():
```
webknjaz, line 205: Alternatively, just do
```suggestion
        for name, min_version in (required_collections or {}).items():
```

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+    if len(role_namespace) == 0:
+        role_namespace = galaxy_infos.get('author', "")
+    # if there's a space in the name space, it's likely author name
+    # and not the galaxy login, so act as if there was no namespace
+    if re.match(r"^\w+ \w+", role_namespace):
+        role_namespace = ""
+    else:
+        role_namespace = f"{role_namespace}."
+    if not isinstance(role_namespace, str):
+        raise RuntimeError("Role namespace must be string, not %s" % role_namespace)
```
webknjaz, line 230: This doesn't make sense. It's always set to string right above.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+    yaml = yaml_from_file("meta/main.yml")
+    if 'galaxy_info' not in yaml:
+        return
+
+    fqrn = _get_role_fqrn(yaml['galaxy_info'])
+
+    if 'role-name' not in options.skip_list:
+        if not re.match(r"[a-z0-9][a-z0-9_]+\.[a-z][a-z0-9_]+$", fqrn):
+            msg = (
+                """\
```
webknjaz, line 263: I'd be nice to put this huge thing into a module-level constant as a template and render it here via `str.format()`.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+            path_list.append(path)
+
+    _update_env('ANSIBLE_LIBRARY', library_paths)
+    _update_env(ansible_collections_path(), collection_list)
+    _update_env('ANSIBLE_ROLES_PATH', roles_path, default=ANSIBLE_DEFAULT_ROLES_PATH)
+
+
+def _update_env(varname: str, value: List[str], default: str = "") -> None:
+    """Update colon based environment variable if needed. by appending."""
+    if value:
```
webknjaz, line 329: Use a guard expression here.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+            return result.groups()[0]
+    elif kind == list:
+        result = re.search(rf"^{key}.* = (\[.*\])$", config, re.MULTILINE)
+        if result:
+            val = eval(result.groups()[0])  # pylint: disable=eval-used
+            if not isinstance(val, list):
+                raise RuntimeError(f"Unexpected data read for {key}: {val}")
+            return val
+    else:
+        raise RuntimeError("Unknown data type.")
```
webknjaz, line 366: Sounds like a `LookupError` would be more appropriate here.

src/ansible_compat/prerun.py
webknjaz: also, it's already being accessed via `packaging.version()` below anyway.

src/ansible_compat/prerun.py
```diff
@@ -0,0 +1,427 @@
...
+        sys.exit(f"Unable to determine ansible collection paths. ({paths})")
+
+    if options.cache_dir:
+        # if we have a cache dir, we want to be use that would be preferred
+        # destination when installing a missing collection
+        paths.insert(0, f"{options.cache_dir}/collections")
+
+    for path in paths:
+        collpath = os.path.join(path, 'ansible_collections', ns, coll)
+        if os.path.exists(collpath):
```
webknjaz, line 394: this loop has a huge potential of using guard expressions

src/ansible_compat/config.py
ssbarnea: If you are talking about https://www.python.org/dev/peps/pep-0562/  that is py37+, so not usable for us.

I am aware of this issue and I plan to refector the library to use a manager that can return  property. That would be very useful for other bits inside too. Still, for the moment I will keep the original implementation in this PR.

src/ansible_compat/constants.py
ssbarnea: Is ok, we resolved it ourselves.

src/ansible_compat/prerun.py
ssbarnea: Added to https://github.com/ansible-community/ansible-compat/issues/6

src/ansible_compat/prerun.py
ssbarnea: Excuse my ignorance but could you show me how it should look like? (maybe even include a link an article).

src/ansible_compat/prerun.py
ssbarnea: I would prefer to keep it like this as it assures that errors are nicely displayed by the consumers. Other exceptions may produce undesired stacktraces. We can debate that later but is better to avoid breaking the current "api".

src/ansible_compat/prerun.py
ssbarnea: Added to https://github.com/ansible-community/ansible-compat/issues/6

src/ansible_compat/config.py
webknjaz: There's a backport: https://pypi.org/project/pep562/

src/ansible_compat/prerun.py
webknjaz: Guard expressions are simple conditionals that interrupt the control flow early. Like "if X: return" or "if Y: continue/break" (for loops).

Typically, when you have something like:
```python
if True:
    a huge
    indented block
    of code
```

you could dedent that block of code if you just invert that if-conditional expression:
```python
if not True:
    return

a huge
indented block
of code
```

And so this refactoring approach is called "guard expression". It allows you have less highly nested code.

src/ansible_compat/prerun.py
webknjaz: This exception is too generic. It's almost the same as just using `raise Exception`. You just don't do this.
Besides, you're leaking the implementation details from other abstraction layers.
Normally, you need to communicate exceptions that make sense for this function and not some wrapper that may or may not be there.
Ideally, you have to raise precise narrow exceptions and convert them to other exceptions at outer levels when it makes sense. It's something the caller has enough context to decide upon, not a low-level lib deep in the call stack.

webknjaz: If you want compat with the old API, this could be an optional layer in this repo that some project may temporarily opt-in but it's mustn't be the default because you're shadowing what's happening. Otherwise, you make a dependency on the CLI of other projects while the dependency direction should be different. We shouldn't have cyclic deps.

PR #8, opened 2021-06-25
Add runtime manager class

ssbarnea:
Initial feedback should answer:

* Is `Runtime()` class ok or you have a better proposal?
* Does the README example look at simple a possible to use?
* Runtime construction accepts two optional arguments
   * project_dir, defaulting to cwd
   * isolated bool, which determine if we should make use of our isolated cache for for installing roles or collections

While current minimal implementation only calls ansible from command line, it will be possible to use the same interface to run ansible execution environments in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,30 @@
 # ansible-compat
 
-A python package contains functions that facilitates working with various
-versions of Ansible, 2.9 and newer.
+A python package contains functions that facilitate working with various
+versions of Ansible 2.9 and newer.
+
+## Using Ansible runtime
+
+```python
+from ansible_compat.runtime import Runtime
+
+def test_runtime():
+
+    # instantiate the runtime using isolated mode, so installing new
+    # roles/collections do not pollute the default setup.
+    runtime = Runtime(isolated=True)
+
+    # Print Ansible core version
+    print(runtime.version)  # 2.9.10 (Version object)
+    # Get configuration info from runtime
+    print(runtime.config.collections_path)
+
+    # Install a new collection
+    # runtime.install_collection("community.general")
+
+    # Execute a command
+    result = runtime.exec(["ansible-doc", "--list"])
+```
 
 ## Access to Ansible configuration
 
--- a/src/ansible_compat/errors.py
+++ b/src/ansible_compat/errors.py
@@ -1,5 +1,10 @@
 """Module to deal with errors."""
+from typing import TYPE_CHECKING, Any, Optional
+
 from ansible_compat.constants import ANSIBLE_MISSING_RC, INVALID_PREREQUISITES_RC
+
+if TYPE_CHECKING:
+    from subprocess import CompletedProcess
 
 
 class AnsibleCompatError(RuntimeError):
@@ -7,11 +12,37 @@ class AnsibleCompatError(RuntimeError):
 
     code = 1  # generic error
 
+    def __init__(
+        self, message: Optional[str] = None, proc: Optional[Any] = None
+    ) -> None:
+        """Construct generic library exception."""
+        super().__init__(message)
+        self.proc = proc
+
+
+class AnsibleCommandError(RuntimeError):
+    """Exception running an Ansible command."""
+
+    def __init__(self, proc: "CompletedProcess[Any]") -> None:
+        """Construct an exception given a completed process."""
+        message = (
+            f"Got {proc.returncode} exit code while running: {' '.join(proc.args)}"
+        )
+        super().__init__(message)
+        self.proc = proc
+
 
 class MissingAnsibleError(AnsibleCompatError):
     """Reports a missing or broken Ansible installation."""
 
     code = ANSIBLE_MISSING_RC
+
+    def __init__(
+        self, message: Optional[str] = None, proc: Optional[Any] = None
+    ) -> None:
+        """."""
+        super().__init__(message)
+        self.proc = proc
 
 
 class InvalidPrerequisiteError(AnsibleCompatError):
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -1,0 +1,89 @@
+"""Ansible runtime environment maanger."""
+import os
+import subprocess
+from typing import TYPE_CHECKING, Any, List, Optional, Union
+
+from packaging.version import Version
+
+from ansible_compat.config import AnsibleConfig, parse_ansible_version
+from ansible_compat.errors import MissingAnsibleError
+from ansible_compat.prerun import get_cache_dir
+
+if TYPE_CHECKING:
+    # https://github.com/PyCQA/pylint/issues/3240
+    # pylint: disable=unsubscriptable-object
+    CompletedProcess = subprocess.CompletedProcess[Any]
+else:
+    CompletedProcess = subprocess.CompletedProcess
+
+
+class Runtime:
+    """Ansible Runtime manager."""
+
+    _version: Optional[Version] = None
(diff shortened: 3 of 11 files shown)


src/ansible_compat/runtime.py
```diff
@@ -0,0 +1,89 @@
+"""Ansible runtime environment maanger."""
```
webknjaz, line 1: s/aa/a/

ssbarnea: is not a mistake is on purpose! See https://en.wikipedia.org/wiki/Manger ;)

webknjaz: But the wiki page shows a word with just one "a" :)

README.md
```diff
@@ -1,7 +1,30 @@
...
+    # Print Ansible core version
+    print(runtime.version)  # 2.9.10 (Version object)
+    # Get configuration info from runtime
+    print(runtime.config.collections_path)
+
+    # Install a new collection
+    # runtime.install_collection("community.general")
+
+    # Execute a command
+    result = runtime.exec(["ansible-doc", "--list"])
```
webknjaz, line 26: I wouldn't call the method `exec()` because it creates a misleading impression that it is related to the syscall that replaces the current process with a new one.

src/ansible_compat/errors.py
```diff
@@ -1,18 +1,49 @@
...
 
+    def __init__(
+        self, message: Optional[str] = None, proc: Optional[Any] = None
+    ) -> None:
+        """Construct generic library exception."""
+        super().__init__(message)
+        self.proc = proc
+
+
+class AnsibleCommandError(RuntimeError):
```
webknjaz, line 23: It'd be useful to have a common lib-defined base class for all the exceptions.

webknjaz: Additionally, this should also inherit `subprocess.SubprocessError`

src/ansible_compat/errors.py
```diff
@@ -1,18 +1,49 @@
...
 
 class MissingAnsibleError(AnsibleCompatError):
     """Reports a missing or broken Ansible installation."""
 
     code = ANSIBLE_MISSING_RC
 
+    def __init__(
+        self, message: Optional[str] = None, proc: Optional[Any] = None
+    ) -> None:
+        """."""
```
webknjaz, line 43: ?

src/ansible_compat/errors.py
webknjaz: Actually, I'm not even sure if it should depend on `RuntimeError` because its purpose is to indicate unspecific unexpected things happening at runtime.

src/ansible_compat/runtime.py
```diff
@@ -0,0 +1,89 @@
...
+        does not affect Ansible installation, an unique cache directory
+        being used instead.
+        """
+        self.project_dir = project_dir or os.getcwd()
+        if isolated:
+            self.cache_dir = get_cache_dir(self.project_dir)
+        self.config = AnsibleConfig()
+
+    # pylint: disable=no-self-use
+    def exec(self, args: Union[str, List[str]]) -> CompletedProcess:
```
webknjaz, line 41: I'd rather have a `check_call()` style thingy. Plus we need to address the naming concern I  raised above.

src/ansible_compat/runtime.py
```diff
@@ -0,0 +1,89 @@
...
+
+        If version is not mentioned, it returns current version as detected.
+        When version argument is mentioned, it return converts the version string
+        to Version object in order to make it usable in comparisons.
+        """
+        if self._version:
+            return self._version
+
+        proc = self.exec(["ansible", "--version"])
+        if proc.returncode == 0:
```
webknjaz, line 63: Having `if .. != 0: raise` could simplify this block.

test/test_runtime.py
```diff
@@ -0,0 +1,28 @@
...
+def test_runtime_version() -> None:
+    """Tests version property."""
+    runtime = Runtime()
+    version = runtime.version
+    assert isinstance(version, Version)
+    # tests that caching property value worked (coverage)
+    assert version == runtime.version
+
+
+def test_runtime_version_fail(mocker: MockerFixture) -> None:
```
webknjaz, line 18: Nit: I normally advocate for using pytest's builtin `monkeypatch` for as long as you don't use things uniquely accessible only via `unittest.mock.patch`.

test/test_runtime.py
```diff
@@ -0,0 +1,28 @@
...
+    """Tests for failure to detect Ansible version."""
+    mocker.patch(
+        "ansible_compat.runtime.parse_ansible_version",
+        return_value=("", "some error"),
+        autospec=True,
+    )
+    runtime = Runtime()
+    with pytest.raises(RuntimeError) as exc:
+        _ = runtime.version
```
webknjaz, line 27: `_ = ` part isn't actually needed here — you could just test accessing the property and that would work.

test/test_runtime.py
```diff
@@ -0,0 +1,28 @@
...
+
+def test_runtime_version_fail(mocker: MockerFixture) -> None:
+    """Tests for failure to detect Ansible version."""
+    mocker.patch(
+        "ansible_compat.runtime.parse_ansible_version",
+        return_value=("", "some error"),
+        autospec=True,
+    )
+    runtime = Runtime()
+    with pytest.raises(RuntimeError) as exc:
```
webknjaz, line 26: Looks like this should've been `MissingAnsibleError` and the idiom is to always use `match=` arg here.

README.md
ssbarnea: Suggest a name, I am aware of its bad naming. I considered `Runtime.run()` as it will likely be very similar to subprocess.run, but ...

src/ansible_compat/errors.py
ssbarnea: Clearly not SubprocessError because we may endup executing code with something else (containers) but I am open to suggestions.

src/ansible_compat/errors.py
ssbarnea: Boilerplate to pass linting, is means "TODO: add meaningful description here." ;)

test/test_runtime.py
ssbarnea: `_` was added to keep the linter happy, lost of ugly things goes towards passing our linters.

src/ansible_compat/errors.py
webknjaz: When you execute things in containers, you still run subprocessed. It's a good abstraction.

src/ansible_compat/errors.py
webknjaz: Yeah, but why not fill it in right away?

test/test_runtime.py
webknjaz: I'd prefer noqa/disable comments because they communicate the intent more clearly.

src/ansible_compat/runtime.py
```diff
@@ -0,0 +1,89 @@
...
+
+if TYPE_CHECKING:
+    # https://github.com/PyCQA/pylint/issues/3240
+    # pylint: disable=unsubscriptable-object
+    CompletedProcess = subprocess.CompletedProcess[Any]
+else:
+    CompletedProcess = subprocess.CompletedProcess
+
+
+class Runtime:
```
webknjaz, line 20: I think this may need to implement a CM at some point.

README.md
webknjaz: I commented below but it should instead just follow `check_call()` style.

PR #9, opened 2021-06-28
Make ansible_config_get raise an error when configuration item is missing

ssbarnea:
- change implementation to raise KeyError by default for missing keys
- allow setting a default value, including None
- this would count as major change but because we do not have any release, it should only worth mentioning that in order to keep old behavior user would have to call it with `default=None`.

Follow-Up: https://github.com/ansible-community/ansible-compat/pull/5#discussion_r659281990

### DIFF
--- a/src/ansible_compat/prerun.py
+++ b/src/ansible_compat/prerun.py
@@ -28,6 +28,7 @@ from ansible_compat.constants import (  # INVALID_CONFIG_RC,
 from ansible_compat.loaders import yaml_from_file
 
 _logger = logging.getLogger(__name__)
+SENTINEL = object()
 
 
 def check_ansible_presence(exit_on_error: bool = False) -> Tuple[str, str]:
@@ -335,7 +336,9 @@ def _update_env(varname: str, value: List[str], default: str = "") -> None:
             _logger.info("Added %s=%s", varname, value_str)
 
 
-def ansible_config_get(key: str, kind: Type[Any] = str) -> Union[str, List[str], None]:
+def ansible_config_get(
+    key: str, kind: Type[Any] = str, default: object = SENTINEL
+) -> Union[str, List[str], None, object]:
     """Return configuration item from ansible config."""
     env = os.environ.copy()
     # Avoid possible ANSI garbage
@@ -361,8 +364,11 @@ def ansible_config_get(key: str, kind: Type[Any] = str) -> Union[str, List[str],
                 raise RuntimeError(f"Unexpected data read for {key}: {val}")
             return val
     else:
-        raise RuntimeError("Unknown data type.")
-    return None
+        raise NotImplementedError("Unknown data type %s." % kind)
+
+    if default == SENTINEL:
+        raise KeyError(key)
+    return default
 
 
 def require_collection(  # noqa: C901
--- a/test/test_prerun.py
+++ b/test/test_prerun.py
@@ -213,6 +213,27 @@ def test_ansible_config_get() -> None:
     assert len(paths) > 0
 
 
+@pytest.mark.parametrize(
+    "default",
+    (
+        (None,),
+        (123,),
+    ),
+)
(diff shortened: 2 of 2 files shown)


tadeboro (changes requested):
The commit message and the implementation do not match, so this cannot be merged as-is. Does setting a default value mean that if the configuration option is not found, we return back a default value?

ssbarnea:
> The commit message and the implementation do not match, so this cannot be merged as-is. Does setting a default value mean that if the configuration option is not found, we return back a default value?

Yep, that is how default works, aims to match python `dict.get()` behavior.

tadeboro:
> Yep, that is how default works, aims to match python `dict.get()` behavior.

If I understand the code correctly, we fail to do this since the default value is ignored if it is not set to a sentinel value. So no matter what default value the caller provides, we always return None if the config is not there.

ssbarnea:
> The commit message and the implementation do not match, so this cannot be merged as-is. Does setting a default value mean that if the configuration option is not found, we return back a default value?

Unless I did an mistake I did not see the purpose of this change is to make use of the [sentinel-object pattern](https://python-patterns.guide/python/sentinel-object/) in order to allow a default value that is not None, so the code can distinguish between being called without mentioning default parameter or being called with `default=None`.

In fact there is even a PEP that proposed introduction of a standard sentinel object in python, but it was not yet approved.

@webknjaz Can you please double check that I did it correctly?

tadeboro:
We clearly need tests for this, but in the meantime, let us have a look at the following call:

    ansible_config_get("key", str, "fallback")

There are two options now:

 1. If the `key` is present in the Ansible output, we return back the associated value. This is OK and expected.
 2. If the `key` is not present in the Ansible output, we return back `None`. And this is not OK since we expect to get back our `fallback` value.

Roughly the same thing also holds for other parameter kinds, but you get the idea.

ssbarnea:
Sorry, i did not see `return None` the intent was `return default` :p

ssbarnea:
In fact I do have the impression that we will remove this method very soon. I already have a [PR open](https://github.com/ansible-community/ansible-compat/pull/10) to add a more flexible config class, that one does not really scale.

tadeboro (approved):

src/ansible_compat/prerun.py
```diff
@@ -361,8 +364,11 @@ def ansible_config_get(key: str, kind: Type[Any] = str) -> Union[str, List[str],
                 raise RuntimeError(f"Unexpected data read for {key}: {val}")
             return val
     else:
-        raise RuntimeError("Unknown data type.")
-    return None
+        raise NotImplementedError("Unknown data type %s." % kind)
+
+    if default == SENTINEL:
```
webknjaz, line 369: @ssbarnea the idiom is the same as with `None` — you must check for the identity, not value:
```suggestion
    if default is SENTINEL:
```

test/test_prerun.py
```diff
@@ -213,6 +213,27 @@ def test_ansible_config_get() -> None:
...
+    result = prerun.ansible_config_get("NON_EXISTING_OPTION", default=default)
+    assert result is default
+
+
+def test_ansible_config_get_raise() -> None:
+    """Check that config get raise if key is not found."""
+    key = "NON_EXISTING_OPTION"
+    with pytest.raises(KeyError) as exc:
+        prerun.ansible_config_get(key)
+    assert exc.value.args[0] == key
```
webknjaz, line 234: The correct idiom is to use `match=` arg in `pytest.raises()` and not check it separately.

ssbarnea: I will keep that in mind, is much cleaner. Anyway, that file was removed so the fix is no longer needed.

webknjaz: Okay. AFAIK ansible/ansible avoided `match=` for some time because of a bug in old pytest versions but in general, this is the right way.

PR #10, opened 2021-06-28
Add class for accessing Ansible configuration

ssbarnea:
This enables easy access to Ansible configuration.

### DIFF
--- a/src/ansible_compat/config.py
+++ b/src/ansible_compat/config.py
@@ -1,14 +1,22 @@
 """Store configuration options as a singleton."""
+import ast
 import os
 import re
 import subprocess
 import sys
+from collections import UserDict
 from functools import lru_cache
-from typing import List, Optional, Tuple
+from typing import TYPE_CHECKING, List, Optional, Tuple
 
 from packaging.version import Version
 
 from ansible_compat.constants import ANSIBLE_MISSING_RC
+
+if TYPE_CHECKING:
+    # https://github.com/PyCQA/pylint/issues/3285
+    _UserDict = UserDict[str, object]  # pylint: disable=unsubscriptable-object
+else:
+    _UserDict = UserDict
 
 # Used to store collection list paths (with mock paths if needed)
 collection_list: List[str] = []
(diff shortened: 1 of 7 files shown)


ssbarnea:
@webknjaz Any change to give me hand with error below:
```
src/ansible_compat/config.py:81: error: Missing type parameters for generic type "UserDict"
```

I am not really sure how to subclass UserDict and keep mypy happy.... I did google a lot but still no clear solution yet.

src/ansible_compat/config.py
```diff
@@ -77,5 +84,50 @@ def ansible_version(version: str = "") -> Version:
...
+        config = subprocess.check_output(
+            ["ansible-config", "dump"], universal_newlines=True, env=env
+        )
+        for match in re.finditer(
+            r"^(?P<key>[A-Za-z0-9_]+).* = (?P<value>.*)$", config, re.MULTILINE
+        ):
+            key = match.groupdict()['key']
+            value = match.groupdict()['value']
+            try:
+                self[key] = eval(value)  # pylint: disable=eval-used
```
felixfontein, line 114: I would really avoid using `eval` in anything than testing code.

src/ansible_compat/config.py
```diff
@@ -77,5 +84,50 @@ def ansible_version(version: str = "") -> Version:
...
+
+    def __init__(self) -> None:
+        """Load config dictionary."""
+        super().__init__()
+        env = os.environ.copy()
+        # Avoid possible ANSI garbage
+        env["ANSIBLE_FORCE_COLOR"] = "0"
+
+        config = subprocess.check_output(
+            ["ansible-config", "dump"], universal_newlines=True, env=env
```
felixfontein, line 106: How about doing this differently:
1. Have a class `AnsibleConfig` which takes the stdout of `ansible-config dump`.
2. Have a separate function that calls `ansible-config dump` and creates `AnsibleConfig` from the result.

This makes it possible to use this class in situations where you want to execute `ansible-config` in a specific venv, or want to execute it asynchronously (and not with `subprocess`).

src/ansible_compat/config.py
ssbarnea: Believe me that I did not want to use eval either but I did not know exactly what to use. I am open to suggestions here.

ssbarnea: I added #11 for tracking this bit in follow-up.

src/ansible_compat/config.py
ssbarnea: In fact I was planning to do this in follow-up, but now that I mentioned it I just did it. Note that I did not split yet the call to ansible, we will do this later with some king of ansible-runner class that will allow us to run various ansible commands, not only galaxy install.

ssbarnea:
Please review, I think i sorted all comments.

src/ansible_compat/config.py
webknjaz: I'd also add that this "function" might as well be a `@classmethod` — it's a common tool for custom initializers.

PR #14, opened 2021-06-30
Enable code coverage

ssbarnea:


test/__init__.py
```diff
@@ -0,0 +1 @@
+"""Tests for ansible_compat package."""
```
webknjaz, line 1: This file shouldn't be necessary.

PR #16, opened 2021-07-01
Move prerun functions into Runtime class

ssbarnea:


src/ansible_compat/config.py
```diff
@@ -95,10 +93,18 @@ class AnsibleConfig(_UserDict):  # pylint: disable=too-many-ancestors
         'COLLECTIONS_PATH': 'COLLECTIONS_PATHS',  # 2.10+ -> 2.9
     }
 
-    def __init__(self, config_dump: Optional[str] = None) -> None:
+    def __init__(
+        self,
+        config_dump: Optional[str] = None,
+        data: Optional[Dict[str, object]] = None,
```
webknjaz, line 99: Why `object` and not `Any`?

src/ansible_compat/config.py
```diff
@@ -95,10 +93,18 @@ class AnsibleConfig(_UserDict):  # pylint: disable=too-many-ancestors
         'COLLECTIONS_PATH': 'COLLECTIONS_PATHS',  # 2.10+ -> 2.9
     }
 
-    def __init__(self, config_dump: Optional[str] = None) -> None:
+    def __init__(
+        self,
+        config_dump: Optional[str] = None,
```
webknjaz, line 98: Multiple ways of initializing the struct in the same initializer looks weird and mixes up multiple unrelated abstractions. It's best to do such things in custom classmethod initializers.

Also, I'd avoid any I/O in the default initializer for a similar reason. The main one should just accept the data and assign it internally.

src/ansible_compat/runtime.py
```diff
@@ -68,22 +96,301 @@ def version(self) -> Version:
...
+
+        _logger.info("Running %s", " ".join(cmd))
+        run = subprocess.run(
+            cmd,
+            universal_newlines=True,
+            check=False,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+        )
+        if run.returncode != 0:
```
webknjaz, line 127: I'd not use this C-style obligatory rc check idiom. It's unpythonic and error-prone because it adds extra cognitive load on the users who have to always remember to do this check.
It's better to have `check=True` because this way (1) it forces you to do try/except and you can't just forget to handle an error + (2) you could explicitly chain the exception cause with `raise X from Y` which greatly improves the DX, specifically the respectability of tracebacks.

src/ansible_compat/runtime.py
```diff
@@ -68,22 +96,301 @@ def version(self) -> Version:
...
+        """Install dependencies from a requirements.yml."""
+        if not os.path.exists(requirement):
+            return
+
+        cmd = [
+            "ansible-galaxy",
+            "role",
+            "install",
+            "--roles-path",
+            f"{self.cache_dir}/roles",
```
webknjaz, line 147: It's usually best to couple options and their values on the same line because otherwise, they look standalone which adds more cognitive complexity for the reader to figure out which ones are related and separate those that aren't.

src/ansible_compat/runtime.py
```diff
@@ -68,22 +96,301 @@ def version(self) -> Version:
...
+        offline: bool = False,
+        required_collections: Optional[Dict[str, str]] = None,
+    ) -> None:
+        """Make dependencies available if needed."""
+        if required_collections is None:
+            required_collections = {}
+
+        if not offline:
+            self.install_requirements("requirements.yml")
+            for req in pathlib.Path(".").glob("molecule/*/requirements.yml"):
```
webknjaz, line 200: Looks like there's an abstraction leak here: compat shouldn't "know" about molecule or lint.

PR #18, opened 2021-07-08
Improve pytest raise check

ssbarnea:


test/test_runtime.py
```diff
@@ -34,9 +34,8 @@ def test_runtime_version_fail(mocker: MockerFixture) -> None:
         autospec=True,
     )
     runtime = Runtime()
-    with pytest.raises(RuntimeError) as exc:
+    with pytest.raises(RuntimeError, match="some error"):
         _ = runtime.version
```
webknjaz, line 38: Unrelated, but it'd be clearer if there' was a "noqa" instead of a magic var.

webknjaz (approved):

PR #19, opened 2021-07-08
Make dependency on tenacity soft

ssbarnea:


src/ansible_compat/_compat.py
```diff
@@ -0,0 +1,31 @@
...
+import logging
+import sys
+from typing import Callable, TypeVar
+
+try:
+    import tenacity
+except ImportError:
+    pass
+
+WrappedFn = TypeVar("WrappedFn", bound=Callable)
```
ssbarnea, line 11: @webknjaz Do you have an idea how to fix the mypy error on Callable. I tried so many approaches, now I need a break.

setup.cfg
```diff
@@ -72,6 +71,8 @@ test =
   pytest-markdown
   pytest-mock
   pytest-plus
+tenacity =
```
webknjaz, line 74: maybe call something like
```suggestion
retry =
```

src/ansible_compat/_compat.py
```diff
@@ -0,0 +1,31 @@
+"""Internal glue."""
```
webknjaz, line 1:
```suggestion
"""Internal compatibility shims."""
```

constraints.txt
```diff
@@ -38,9 +38,5 @@ pytest-plus==0.2
     # via ansible-compat (setup.py)
 pyyaml==5.4.1
     # via ansible-compat (setup.py)
-six==1.16.0
-    # via tenacity
-tenacity==7.0.0
```
webknjaz, line 43: I don't think this should be removed from the pins.

src/ansible_compat/_compat.py
```diff
@@ -0,0 +1,31 @@
...
+WrappedFn = TypeVar("WrappedFn", bound=Callable)
+_logger = logging.getLogger(__name__)
+
+
+def noop_dec(f: WrappedFn) -> WrappedFn:
+    """Decorate function with a no-op."""
+    return f
+
+
+if "tenacity" in sys.modules:
```
webknjaz, line 20: Plz set a var during the import attempt and use it then. It's better than doing a list scan in the import-time (just like `'tenacity' in globals()`)

src/ansible_compat/_compat.py
```diff
@@ -0,0 +1,31 @@
...
+        reraise=True,
+        wait=tenacity.wait_fixed(30),  # type: ignore
+        stop=tenacity.stop_after_attempt(3),  # type: ignore
+        before_sleep=tenacity.after_log(_logger, logging.WARNING),  # type: ignore
+    )
+else:
+    retry = noop_dec
+
+
+__all__ = ["retry"]
```
webknjaz, line 31: This must be immutable. Normally, pylint warns about this but that check seems to be disabled for some reason.
```suggestion
__all__ = "retry",
```

src/ansible_compat/_compat.py
```diff
@@ -0,0 +1,31 @@
...
+
+if "tenacity" in sys.modules:
+    retry = tenacity.retry(  # Retry up to 3 times as galaxy server can return errors
+        reraise=True,
+        wait=tenacity.wait_fixed(30),  # type: ignore
+        stop=tenacity.stop_after_attempt(3),  # type: ignore
+        before_sleep=tenacity.after_log(_logger, logging.WARNING),  # type: ignore
+    )
+else:
+    retry = noop_dec
```
webknjaz, line 28: Put this before "if" and there's no need for "else" then.

src/ansible_compat/_compat.py
```diff
@@ -0,0 +1,31 @@
...
+try:
+    import tenacity
+except ImportError:
+    pass
+
+WrappedFn = TypeVar("WrappedFn", bound=Callable)
+_logger = logging.getLogger(__name__)
+
+
+def noop_dec(f: WrappedFn) -> WrappedFn:
```
webknjaz, line 15: Shouldn't this just use `T`?

src/ansible_compat/_compat.py
webknjaz: Maybe
```suggestion
WrappedFn = TypeVar("WrappedFn", bound=Callable[Any, Any])
```
?

src/ansible_compat/_compat.py
```diff
@@ -0,0 +1,31 @@
...
+    """Decorate function with a no-op."""
+    return f
+
+
+if "tenacity" in sys.modules:
+    retry = tenacity.retry(  # Retry up to 3 times as galaxy server can return errors
```
webknjaz, line 21: Remove the mention of galaxy. It doesn't make any sense on the compat layer.

src/ansible_compat/runtime.py
```diff
@@ -128,12 +128,7 @@ def install_collection(
             _logger.error("Command returned %s code:\n%s", run.returncode, run.stdout)
             raise InvalidPrerequisiteError()
 
-    @tenacity.retry(  # Retry up to 3 times as galaxy server can return errors
-        reraise=True,
-        wait=tenacity.wait_fixed(30),  # type: ignore
-        stop=tenacity.stop_after_attempt(3),  # type: ignore
-        before_sleep=tenacity.after_log(_logger, logging.WARNING),  # type: ignore
-    )
+    @retry
```
webknjaz, line 131: Btw the shim could just accept all these args and bypass them conditionally. Not sure if it's important right now. It's probably okay to keep it if this is the only use atm.

constraints.txt
```diff
@@ -16,7 +16,7 @@ iniconfig==1.1.1
     # via pytest
 more-itertools==8.8.0
     # via pytest-plus
-packaging==20.9
+packaging==21.0
```
webknjaz, line 19: Maybe don't couple this change with the lib removal?

setup.cfg
ssbarnea: I was considering this but I think that we may want to have a basic retry implementation as default instead of noop. In that case having an extra named retry is misleading as it would be another type of retry.

Another issue that worries me a little bit is inability to disable/customize retry behavior. What if consumer of the library does not want any retries? In its current form they are out of options, in fact the behavior would change based on the presence of absence of tenacity, creating some divergence in behavior.

I think that ideally we should add a parameter to the Runtime() class constructor that allows customization of the retry behavior for all the code that may need it. 

Do you have an idea on how we can achieve this level of flexibility in a clean way?

webknjaz: Accept a decorator callable?

src/ansible_compat/_compat.py
ssbarnea: I did not had a problem with original approach but the linters did, especially mypy which was upset about tenacity being either None or a ModuleType, feel free to push to the branch with a fix. As far is passing I don't mind. I will stop doing forced pushes in order to be able to collab on it.

webknjaz: Okay, then just add an explanation comment for anyone having a bright idea to refactor this in the future.

PR #20, opened 2021-07-09
Improve use of pip-compile

ssbarnea:
* By default during linting pip-compile will run and check if deps are ok without attempting to upgrade them
* User can manually call `tox -e deps` which enables the manual hook that does run pip-compile with --upgrade. Basically deps can be used to bump dependencies when desired.


webknjaz (approved):

tox.ini
```diff
@@ -73,11 +71,9 @@ description = Bump all test depeendencies
 basepython = python3
 deps =
   pre-commit>=2.6.0
-  pip-tools>=5.4.0
   setuptools>=51.1.1
 skip_install = true
 commands =
-  pip-compile --upgrade --extra test --output-file=constraints.txt setup.py
   {envpython} -m pre_commit run {posargs:--all-files --hook-stage manual -v}
```
webknjaz, line 77: This should also show the diff on failure.

PR #22, opened 2021-07-19
Add sphinx documentation

ssbarnea:


ssbarnea:
@webknjaz For some reason autodoc fails to include the constructor documentation for Runtime. Instead it does include the class docstring, which does not include info about various arguments.

docs/conf.py
```diff
@@ -0,0 +1,94 @@
...
+# Usually you set "language" from the command line for these cases.
+language = None
+
+# List of patterns, relative to source directory, that match files and
+# directories to ignore when looking for source files.
+# This pattern also affects html_static_path and html_extra_path .
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+
+# The name of the Pygments (syntax highlighting) style to use.
+pygments_style = 'sphinx'
```
webknjaz, line 72: Drop this or use `ansible`

docs/conf.py
```diff
@@ -0,0 +1,94 @@
...
+html_theme = 'sphinx_ansible_theme'
+
+
+# -- Options for HTMLHelp output ---------------------------------------------
+
+# Output file base name for HTML help builder.
+htmlhelp_basename = 'ansible-compat'
+
+
+intersphinx_mapping = {'https://docs.python.org/': None}
```
webknjaz, line 89: Note that this syntax is outdated. Did you use old sphinx to generate this file?

tox.ini
```diff
@@ -107,3 +107,20 @@ description = Run the tests with {basepython} ansible-core 2.11+
 deps =
   {[testenv]deps}
   ansible-core>=2.11
+
+[testenv:docs]
+description = Build docs
+commands =
+    python -m sphinx \
+      -a -n -W \
```
webknjaz, line 115: Don't forget to add `--keep-going`

src/ansible_compat/runtime.py
```diff
@@ -47,13 +47,17 @@ class Runtime:
...
-        self, project_dir: Optional[str] = None, isolated: bool = False
+        self,
+        project_dir: Optional[str] = None,
+        isolated: bool = False,
     ) -> None:
         """Initialize Ansible runtime environment.
 
-        Isolated mode assures that installation of collections or roles
+        :param isolated: Assure that installation of collections or roles
         does not affect Ansible installation, an unique cache directory
```
webknjaz, line 57: Shouldn't the rest of the paragraph be indented?

docs/index.md
```diff
@@ -0,0 +1,17 @@
+# ansible-compat
+
+```{eval-rst}
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
+.. automodule:: ansible_compat.runtime
+.. autoclass:: Runtime
```
webknjaz, line 9: Try this https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-special-members

docs/conf.py
ssbarnea: Got it from ansible-runner,... bit outdated.

webknjaz: Yep, that's probably because the new format exists since Sphinx v1.0 but only a few months ago I noticed that the quickstart script still uses the old one: https://github.com/sphinx-doc/sphinx/issues/8985.
So that is fixed now since Sphinx v4.0: https://github.com/sphinx-doc/sphinx/pull/8986.

docs/conf.py
```diff
@@ -0,0 +1,80 @@
...
+
+# -- Options for HTMLHelp output ---------------------------------------------
+
+# Output file base name for HTML help builder.
+htmlhelp_basename = 'ansible-compat'
+
+
+intersphinx_mapping = {
+    'python': ('https://docs.python.org/3', None),
+    'packaging': ('https://packaging.python.org/en/latest/', None),
```
webknjaz, line 73: FWIW I'd use the 'PyPUG' alias (I think pip calls it like this too), to avoid confusion with the library called 'packaging'.

Issue #6, opened 2021-06-25
Fix problems from original core from ansiblelint.prerun


ssbarnea:
List of problems identified and by whom:

- [ ] check_ansible_presence() return type is not pythonic, @webknjaz 
- [x] do we need tenacity or something else? @webknjaz (https://github.com/ansible-community/ansible-compat/issues/12)
- [x] avoid running code on module import: ansible_collections_path @webknjaz

PR #23, opened 2021-07-20
Add ability to check minimal ansible version

ssbarnea:
This will allow us to remove the old `prerun.check_ansible_presence` implementation and make instantiation quite nice an easy.

Also this approach is compatible with a possible future use of ansible-runner.

Partial-Fix: #6

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -50,6 +50,7 @@ class Runtime:
         self,
         project_dir: Optional[str] = None,
         isolated: bool = False,
+        min_required_version: Optional[str] = None,
     ) -> None:
         """Initialize Ansible runtime environment.
 
@@ -59,12 +60,23 @@ class Runtime:
         :param isolated: Assure that installation of collections or roles
                          does not affect Ansible installation, an unique cache
                          directory being used instead.
+        :param min_required_version: Minimal version of Ansible required. If
+                                     not found, a :class:`RuntimeError`
+                                     exception is raised.
         """
         self.project_dir = project_dir or os.getcwd()
         self.isolated = isolated
         if isolated:
             self.cache_dir = get_cache_dir(self.project_dir)
         self.config = AnsibleConfig()
+
+        if (
+            min_required_version is not None
+            and packaging.version.Version(min_required_version) > self.version
+        ):
+            raise RuntimeError(
+                f"Found incompatible version of ansible runtime {self.version}, instead of {min_required_version} or newer."
+            )
 
     def clean(self) -> None:
         """Remove content of cache_dir."""
--- a/test/test_runtime.py
+++ b/test/test_runtime.py
@@ -26,6 +26,12 @@ def test_runtime_version(runtime: Runtime) -> None:
     assert version == runtime.version
 
 
+def test_runtime_version_outdated() -> None:
+    """Checks that instantiation raises if version is outdated."""
+    with pytest.raises(RuntimeError, match="Found incompatible version of ansible"):
+        Runtime(min_required_version="9999.9.9")
+
+
 def test_runtime_version_fail(mocker: MockerFixture) -> None:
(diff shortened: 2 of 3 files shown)


src/ansible_compat/runtime.py
```diff
@@ -59,13 +60,21 @@ def __init__(
         :param isolated: Assure that installation of collections or roles
                          does not affect Ansible installation, an unique cache
                          directory being used instead.
+        :param version: Minimal version of Ansible required. If not found,
+                        a RuntimError exception is raised.
```
webknjaz, line 64: typo+role
```suggestion
                        a :py:class:`RuntimeError` exception is raised.
```

src/ansible_compat/runtime.py
```diff
@@ -59,13 +60,21 @@ def __init__(
...
         self.project_dir = project_dir or os.getcwd()
         self.isolated = isolated
         if isolated:
             self.cache_dir = get_cache_dir(self.project_dir)
         self.config = AnsibleConfig()
 
+        if version and packaging.version.Version(version) > self.version:
+            raise RuntimeError(
+                "Found incompatible version of ansible runtime %s, instead of %s or newer."
+                % (self.version, version)
```
webknjaz, line 75: We could just use an f-string here, right? If you insist on using placeholders, please make them named (like `%(version)s` for %-formatting or `{version!s}` for `str.format()`).

test/test_runtime.py
```diff
@@ -26,6 +26,12 @@ def test_runtime_version(runtime: Runtime) -> None:
     assert version == runtime.version
 
 
+def test_runtime_version_outdated() -> None:
+    """Checks that instantiation raises if version is outdated."""
+    with pytest.raises(RuntimeError):
```
webknjaz, line 31: Always use `match=` with `pytest.raises` or `pytest.warns`

src/ansible_compat/runtime.py
```diff
@@ -50,6 +50,7 @@ def __init__(
         self,
         project_dir: Optional[str] = None,
         isolated: bool = False,
+        version: Optional[str] = None,
```
webknjaz, line 53: The docstring says that it's a minimal version. Just version may mean "install/upgrade this version if missing". But if you improve the naming, it may become clear what it is:
```suggestion
        min_required_version: Optional[str] = None,
```

src/ansible_compat/runtime.py
```diff
@@ -59,13 +60,21 @@ def __init__(
...
+        :param version: Minimal version of Ansible required. If not found,
+                        a RuntimError exception is raised.
         """
         self.project_dir = project_dir or os.getcwd()
         self.isolated = isolated
         if isolated:
             self.cache_dir = get_cache_dir(self.project_dir)
         self.config = AnsibleConfig()
 
+        if version and packaging.version.Version(version) > self.version:
```
webknjaz, line 72: "if version" is semantically incorrect and may result in an unexpected control flow. Like if it's `0`.
The correct way of the comparison with `None` is to have an identity check like `if version is not None`.

test/test_runtime.py
```diff
@@ -26,6 +26,12 @@ def test_runtime_version(runtime: Runtime) -> None:
     assert version == runtime.version
 
 
+def test_runtime_version_outdated() -> None:
```
webknjaz, line 29: I think this test is a good candidate for `@pytest.mark.parametrize` — this way you'll document which types of values you had in mind when adding this feature.

ssbarnea: I go for parametrize when there is a clear benefit in avoiding duplication, but for cases like this I prefer not using it. The reason behind is that parameterized tests are harder to read and to run as you need to add extra quotes on shell due to the use of `[...]` in their ids. Anyway, we can easily rewrite them when needed.

webknjaz: No need to merge this with other tests. Just add something other than 99, like corner cases.

PR #25, opened 2021-07-20
Add ability to check Ansible python module

ssbarnea:
- add ability to check presence of Ansible python module
- remove pre.check_ansible_presence() as we do not want to support it, Runtime() constructor being its successor.

Related: #6


src/ansible_compat/runtime.py
```diff
@@ -77,6 +83,29 @@ def __init__(
             raise RuntimeError(
                 f"Found incompatible version of ansible runtime {self.version}, instead of {min_required_version} or newer."
             )
+        if require_module:
+            self._check_module()
+
+    def _check_module(self) -> None:
```
webknjaz, line 89:
```suggestion
    def _ensure_module_available(self) -> None:
```

src/ansible_compat/runtime.py
```diff
@@ -77,6 +83,29 @@ def __init__(
...
+        try:
+            # pylint: disable=import-outside-toplevel
+            from ansible.release import __version__
+
+            ansible_module_version = packaging.version.parse(__version__)
+
+        except (ImportError, ModuleNotFoundError):
+            pass
+
+        if not ansible_module_version:
```
webknjaz, line 101:
```suggestion
        if ansible_module_version is None:
```

src/ansible_compat/runtime.py
```diff
@@ -77,6 +83,29 @@ def __init__(
...
+            self._check_module()
+
+    def _check_module(self) -> None:
+        """Assure that Ansible Python module is installed and matching CLI version."""
+        ansible_module_version = None
+        try:
+            # pylint: disable=import-outside-toplevel
+            from ansible.release import __version__
+
+            ansible_module_version = packaging.version.parse(__version__)
```
webknjaz, line 96: Move this to `else:` block. We don't need to catch exceptions coming from `packaging.version`.

test/test_runtime.py
```diff
@@ -25,11 +25,11 @@ def test_runtime_version(runtime: Runtime) -> None:
     # tests that caching property value worked (coverage)
     assert version == runtime.version
 
-
-def test_runtime_version_outdated() -> None:
+@pytest.mark.parametrize("require_module", [True, False])
```
webknjaz, line 28:
```suggestion
@pytest.mark.parametrize(
    "require_module",
    (True, False),
    ids=("module required", "module unrequired"),
)
```

src/ansible_compat/runtime.py
```diff
@@ -77,6 +83,29 @@ def __init__(
...
                 f"Found incompatible version of ansible runtime {self.version}, instead of {min_required_version} or newer."
             )
+        if require_module:
+            self._check_module()
+
+    def _check_module(self) -> None:
+        """Assure that Ansible Python module is installed and matching CLI version."""
+        ansible_module_version = None
+        try:
+            # pylint: disable=import-outside-toplevel
```
webknjaz, line 93: Why is this comment necessary? Normally, this error happens when you have relative imports.

src/ansible_compat/runtime.py
```diff
@@ -77,6 +83,29 @@ def __init__(
...
+            from ansible.release import __version__
```
webknjaz, line 94: For dynamic imports, I'd prefer using importlib. It also makes it easier to inject `ImportError`

src/ansible_compat/runtime.py
```diff
@@ -77,6 +83,29 @@ def __init__(
...
+
+            ansible_module_version = packaging.version.parse(__version__)
+
+        except (ImportError, ModuleNotFoundError):
```
webknjaz, line 98: `ModuleNotFoundError` is a subclass of `ImportError`. There's no point in listing it additionally — it'll never be hit because `ImportError` is first in the tuple and will be matched first:
```suggestion
        except ImportError:
```

test/test_runtime.py
```diff
@@ -25,11 +25,11 @@ def test_runtime_version(runtime: Runtime) -> None:
...
+def test_runtime_version_outdated(require_module: bool) -> None:
     """Checks that instantiation raises if version is outdated."""
     with pytest.raises(RuntimeError, match="Found incompatible version of ansible"):
```
webknjaz, line 31: This error is probably misleading in case there's no module at all.

PR #28, opened 2021-07-21
Reorganize documentation

ssbarnea:
- move examples into standalone file which is included in builded doc
- put link from project readme to documentation site
- include few basic badges in project readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,46 +1,11 @@
 # ansible-compat
 
+[![pypi](https://img.shields.io/pypi/v/ansible-compat.svg)](https://pypi.org/project/ansible-compat/)
+[![docs](https://readthedocs.org/projects/ansible-compat/badge/?version=latest)](https://ansible-compat.readthedocs.io/en/latest/)
+[![gh](https://github.com/ansible-community/ansible-compat/actions/workflows/tox.yml/badge.svg)](https://github.com/ansible-community/ansible-compat/actions/workflows/tox.yml)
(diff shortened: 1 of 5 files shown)


docs/examples.md
```diff
@@ -0,0 +1,41 @@
+### Using Ansible runtime
```
webknjaz, line 1: Just make it a python file and literal—include it instead.

README.md
```diff
@@ -1,46 +1,11 @@
 # ansible-compat
```
webknjaz, line 1: I think I'd want the readme to be partially included into the docs front page to avoid duplication as much as possible.

docs/index.md
```diff
@@ -17,4 +22,5 @@
 * :ref:`genindex`
 * :ref:`modindex`
 * :ref:`search`
+
```
webknjaz, line 25: Why is this necessary?

README.md
ssbarnea: We cannot really include it because for few reasons:
* it would include phrase that does not make much sense, pointing to the location of official documentation (would be kind recursive to display it on the official docs page)
* not sure if displaying badges is desirable in published doc but if you think so, ok
* header indentation/level would mismatch and we would get an error while building

I make it easier to progress on that just push extra fixes to that branch. I will avoid forced pushes. I have no strong feeling around how we organize it.

webknjaz: That's why I said *partially*. There's start-at, start-after and so on. You should only reuse parts that makes sense to reuse.

ssbarnea: There is nothing reusable inside root `README.md` file.

Issue #27, opened 2021-07-21
Ensure full test code coverage on library


ssbarnea:
We should ensure we have full code-coverage for the library before effectively starting to use it in molecule and linter, especially as the code-base is quite small.

ssbarnea:
Closing with note that anyone is welcomed to make a PR to improve current coverage level.

PR #29, opened 2021-07-23
Improve code coverage

ssbarnea:
Partial-Fix: #27

### DIFF
--- a/test/test_runtime.py
+++ b/test/test_runtime.py
@@ -15,7 +15,7 @@ from pytest_mock import MockerFixture
 
 from ansible_compat.constants import INVALID_PREREQUISITES_RC
 from ansible_compat.errors import AnsibleCompatError, InvalidPrerequisiteError
-from ansible_compat.runtime import Runtime, _update_env
+from ansible_compat.runtime import CompletedProcess, Runtime, _update_env
 
 
 def test_runtime_version(runtime: Runtime) -> None:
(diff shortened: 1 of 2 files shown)


test/test_runtime.py
```diff
@@ -71,6 +71,29 @@ def test_runtime_version_fail(mocker: MockerFixture) -> None:
...
+        RuntimeError, match="Unable to find a working copy of ansible executable."
+    ):
+        runtime.version  # pylint: disable=pointless-statement
+
+
+def test_runtime_install_requirements_missing_file() -> None:
+    """Check that missing requirements file is ignored."""
+    # Do not rely on this behavior, it may be removed in the future
+    runtime = Runtime()
+    runtime.install_requirements("/that/does/not/exist")
```
webknjaz, line 94: Where is the assertion here?

ssbarnea: The only thing expected is not to raise, which is implicit. If you want I can assert that it returns None?

webknjaz: I think that a more intuitive API would be to actually raise an exception.

ssbarnea: I agree with you with only mention that probably we want an optional boolean `raise` option for keeping it like this. Main reason for this is to keep usage simple as the most common pattern is "if reqs file exists install them". A missing file means nothing to install. 

As I am still finding bugs while increasing the coverage, I suspect before making the v1 release and adopting it, we should review all the public API and decide if we like them or not.

PR #30, opened 2021-07-24
Fix standalone role bootstrapping

ssbarnea:
- Fixes a bug introduce when we added support for project_dir and we still continued to rely on CWD in few places.
- Adds tests for covering this use-case

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -240,7 +240,9 @@ class Runtime:
 
         if not offline:
             self.install_requirements("requirements.yml")
-            for req in pathlib.Path(".").glob("molecule/*/requirements.yml"):
+            for req in pathlib.Path(self.project_dir).glob(
+                "molecule/*/requirements.yml"
+            ):
                 self.install_requirements(str(req))
 
         for name, min_version in required_collections.items():
@@ -249,9 +251,8 @@ class Runtime:
                 destination=f"{self.cache_dir}/collections" if self.cache_dir else None,
             )
 
-        _install_galaxy_role(self.project_dir)
-        # _perform_mockings()
         self._prepare_ansible_paths()
+        _install_galaxy_role(self.project_dir)
 
     def require_collection(  # noqa: C901
         self,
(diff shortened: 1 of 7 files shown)


tox.ini
```diff
@@ -24,7 +24,7 @@ commands =
   ansible --version
   # We add coverage options but not making them mandatory as we do not want to force
   # pytest users to run coverage when they just want to run a single test with `pytest -k test`
-  coverage run -m pytest {posargs:}
+  coverage run -m pytest -vvvv {posargs:}
```
webknjaz, line 27: The level of verbosity should be contributed by the caller.

src/ansible_compat/runtime.py
```diff
@@ -122,6 +122,7 @@ def exec(self, args: Union[str, List[str]]) -> CompletedProcess:
             check=False,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
+            env=os.environ,
```
webknjaz, line 125: This happens by default. Why did you add it?

pyproject.toml
```diff
@@ -24,7 +24,8 @@ skip-string-normalization = true
 profile = "black"
 
 [tool.pytest.ini_options]
-addopts = "--no-success-flaky-report"
+addopts = "--no-success-flaky-report -vvvv"
```
webknjaz, line 27: This looks out of the scope also it's inconsistent with the comment below.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "test/assets/ansible-role-mysql"]
```
webknjaz, line 1: It'd go for git subtree or put the synching logic into the pytest setup. Making the contributors do this is fragile.

test/test_runtime.py
```diff
@@ -87,6 +87,25 @@ def test_runtime_version_fail_cli(mocker: MockerFixture) -> None:
...
+    runtime = Runtime(isolated=True, project_dir=project_dir)
+    runtime.prepare_environment()
+    for record in caplog.records:
+        assert record.levelname != "CRITICAL"
+    assert "symlink to current repository in order to enable Ansible to find the role" in caplog.text
+    print(caplog.text)
+    # check that role appears as installed now
+    result = runtime.exec(["ansible-galaxy", "list", "-vvv"])
+    # print(result)
+    assert result.returncode == 0, result
```
webknjaz, line 105: No need to print out the `result` like this — just add `--showlocals` to addopts to reveal it.

ssbarnea:
There is a lot of debugging junk i need to remove. I think I found the culprit yesterday, mainly submodules cloning no longer works with checkout v2. Instead I will just create a mini embedded role for testing.

Issue #26, opened 2021-07-21
allow user to disable retry functionality in the library


ssbarnea:
Some users may not want to make use of our embedded retry mechanism used by network-dependent Runtime() methods and we should allow them to decide if they want a retry mechanism or not.

PR #32, opened 2021-07-25
Change retry support implementation

ssbarnea:
- remove dependency on tenacity
- retry support will be disabled by default
- adds 'retry' argument to Runtime constructor to opt-in for retry logic
- use retry only on galaxy install commands by default

Fixes: #26


src/ansible_compat/runtime.py
```diff
@@ -113,16 +119,29 @@ def clean(self) -> None:
...
-            check=False,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-        )
+    def exec(
+        self, args: Union[str, List[str]], retry: bool = False
+    ) -> CompletedProcess:
+        """Execute a command inside an Ansible environment.
+
+        param: retry: Retry network operations on failures.
```
webknjaz, line 127:
```suggestion
        :param retry: Retry network operations on failures.
```

src/ansible_compat/runtime.py
```diff
@@ -113,16 +119,29 @@ def clean(self) -> None:
...
-            stderr=subprocess.PIPE,
-        )
+    def exec(
+        self, args: Union[str, List[str]], retry: bool = False
+    ) -> CompletedProcess:
+        """Execute a command inside an Ansible environment.
+
+        :param retry: Retry network operations on failures.
+        """
+        for _ in range(self.max_retries if retry else 1):
```
webknjaz, line 129: Why not
```suggestion
        for _ in range(self.max_retries if self.retry else 1):
```

src/ansible_compat/runtime.py
```diff
@@ -165,12 +184,14 @@ def install_collection(
         cmd.append(f"{collection}")
 
         _logger.info("Running %s", " ".join(cmd))
-        run = self.exec(cmd)
+        run = self.exec(
+            cmd,
+            retry=self.retry,
```
webknjaz, line 189: It feels weird having to pass this flag to both the class initializer and the individual method.

src/ansible_compat/runtime.py
ssbarnea: You indeed spot an design issue but that is not really the correct fix. The default retry value used by exec should be self.retry but we should still keep it. Users may want to alter the retry behavior for specific commands, even inside the library itself we do this. Mainly galaxy commands should use retry but other operations should not.

Let's check the use cases to get it right from start.

src/ansible_compat/runtime.py
```diff
@@ -112,16 +116,29 @@ def clean(self) -> None:
...
-            check=False,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-        )
+    def exec(
+        self, args: Union[str, List[str]], retry: bool = False
+    ) -> CompletedProcess:
+        """Execute a command inside an Ansible environment.
+
+        :param retry: Retry network operations on failures.
```
webknjaz, line 124: nit: why isn't `args` documented?

ssbarnea: I was not able to find a nice way to document the args/kwargs on run, as we mainly pass them to `subprocess.run` but I agree that we should expose those that matter.

PR #38, opened 2021-07-27
Increase test coverage

ssbarnea:


test/test_runtime.py
```diff
@@ -360,6 +366,26 @@ def test_require_collection_invalid_name(runtime: Runtime) -> None:
...
+    with pytest.raises(
+        InvalidPrerequisiteError, match="Unable to determine ansible collection paths"
+    ):
+        runtime.require_collection("community.molecule")
+
+
+def test_require_collection_preexisting_broken(tmp_path: pathlib.Path) -> None:
+    """Check that require_collection raise with broken pre-existing collection."""
+    runtime = Runtime(isolated=True, project_dir=str(tmp_path))
+    dest_path: str = runtime.config.collections_path[0]  # type: ignore
```
ssbarnea, line 381: @webknjaz apparently even with our type hints added to Config class, mypy still believes that any `Config.attribute` is of type `object`.

I am not pleased about this dirty trick because it means that this would be expected from consumers of this library too. Any idea on how to avoid that?

PR #42, opened 2021-07-30
Ensure we encourage consistent use of config values

ssbarnea:
Replace previous double directional use of aliases for renamed options, raise warnings when deprecated names are used.

### DIFF
--- a/src/ansible_compat/config.py
+++ b/src/ansible_compat/config.py
@@ -5,6 +5,7 @@ import os
 import re
 import subprocess
 import sys
+import warnings
 from collections import UserDict
 from functools import lru_cache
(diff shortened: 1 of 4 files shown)


src/ansible_compat/config.py
```diff
@@ -454,6 +454,10 @@ def __getattr__(self, attr_name: str) -> object:
         if name in self.data:
             return self.data[name]
         if name in self._aliases:
+            warnings.warn(
+                f'Detected deprecated use of {name}, replace it with {self._aliases[name]}',
+                category=DeprecationWarning,
```
webknjaz, line 459: @ssbarnea in such cases, always use `stacklevel=2`.

PR #47, opened 2021-08-01
Patch codecov config

ssbarnea:


webknjaz:
Maybe turn it into a draft?

ssbarnea:
Github does not yet allow undrafting like gerrit.

webknjaz:
> Github does not yet allow undrafting like gerrit.

What do you mean by that?

ssbarnea:
@webknjaz Do you have any idea why `codecov/patch` check is still reporting with this change. My expectation was to see only `codecov/project` one but both are still there. Based on their docs this should disable it.

webknjaz:
No idea

PR #50, opened 2021-08-01
Refactored ansible_version to be pythonic

ssbarnea:


src/ansible_compat/config.py
```diff
@@ -78,18 +78,12 @@ def ansible_version(version: str = "") -> Version:
...
-        if error is not None:
-            print(error)
-            raise MissingAnsibleError()
-    else:
-        print(
-            "Unable to find a working copy of ansible executable.",
-            proc,
+    if proc.returncode != 0:
+        raise MissingAnsibleError(
+            "Unable to find a working copy of ansible executable.", proc=proc
```
webknjaz, line 83: Why did you drop a trailing comma here?

src/ansible_compat/config.py
```diff
@@ -52,12 +52,12 @@ def parse_ansible_version(stdout: str) -> Tuple[str, Optional[str]]:
...
     if match:
-        return match.group("version"), None
+        return Version(match.group("version"))
     # ansible-base 2.10 and Ansible 2.9: 'ansible 2.x.y'
     match = re.search(r"^ansible (?P<version>[^\s]+)", stdout, re.MULTILINE)
     if match:
-        return match.group("version"), None
-    return "", "FATAL: Unable parse ansible cli version: %s" % stdout
+        return Version(match.group("version"))
+    raise AnsibleCompatError("Unable to parse ansible cli version: %s" % stdout)
```
webknjaz, line 60: Compat is too generic. It should be specific inheriting `LookupError`

ssbarnea: How about `InvalidPrerequisiteError` which already exists?

src/ansible_compat/config.py
ssbarnea: I was not me, it was black. Commas are useful only when you have multiple lines so you can easily add new entry or reorder them.

PR #53, opened 2021-08-13
Remove deprecation warnings

ssbarnea:
It seems that our attempt to recommend use of specific
names does trigger runtime warnings from within the library itself,
risking to confuse users.


webknjaz:
How does this happen?

PR #58, opened 2021-08-16
Fix require_collection bug with missing cache directory

ssbarnea:
Previously lack of expanduser() failed to correctly identify existence of ansible user directory. Bug was found while trying
to ensure full coverage on require_collection.


test/test_runtime.py
```diff
@@ -535,3 +535,10 @@ def test_upgrade_collection(runtime_tmp: Runtime) -> None:
         runtime_tmp.require_collection("containers.podman", "1.6.1", install=False)
     # now we really perform the upgrade
     runtime_tmp.require_collection("containers.podman", "1.6.1")
+
+
+def test_require_collection_no_cache_dir() -> None:
+    """Check require_collection without a cache directory."""
+    runtime = Runtime()
+    assert not runtime.cache_dir
+    runtime.require_collection("community.molecule", "0.1.0", install=True)
```
webknjaz, line 544: There should probably be another assertion after this line.

ssbarnea: At this moment we do not have API to access installed collections but once we do we will update the test. Probably you know that ansible 2.9 is not even able to list installed collections.

There are several other tests were I will want to check the same thing and were the only current test is that "it does not raise", which is implicit.

PR #59, opened 2021-08-17
Fix prepare environment without isolation

ssbarnea:
Includes workaround for https://github.com/PyCQA/pylint/issues/4667 bug.

### DIFF
--- a/test/roles/sample4/meta/main.yml
+++ b/test/roles/sample4/meta/main.yml
@@ -1,0 +1,16 @@
+---
+dependencies: []
+
+galaxy_info:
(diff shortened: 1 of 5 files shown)


test/roles/sample4/meta/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+dependencies: []
+
+galaxy_info:
+  # role_name is missing in order to test deduction from folder name
+  author: acme
+  description: ACME samble role
```
webknjaz, line 7: *sample

PR #61, opened 2021-08-17
Make _install_galaxy_role a Runtime method

ssbarnea:
This is is preparatory change for #59 which is also fixing some bugs. There is no behavior change on this refactoring.

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -255,7 +255,7 @@ class Runtime:
 
         self._prepare_ansible_paths()
         # install role if current project looks like a standalone role
-        _install_galaxy_role(self.project_dir, ignore_errors=True)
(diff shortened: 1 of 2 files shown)


src/ansible_compat/runtime.py
```diff
@@ -354,69 +354,69 @@ def _prepare_ansible_paths(self) -> None:
             'ANSIBLE_ROLES_PATH', roles_path, default=ANSIBLE_DEFAULT_ROLES_PATH
         )
 
+    # pylint: disable=no-self-use
```
webknjaz, line 357: Instead of this, you should probably use `self.project_dir` somewhere in code.

ssbarnea: This was there only because I wanted to split a bigger change in two for clarity. The next commit did remove it. It was not meant to stay in for long.

PR #66, opened 2021-09-02
Add version_in_range method to Runtime

ssbarnea:
Adds a new API method that can easily be used to detect if ansible is within a specified range. That should cover the most common use case of Ansible version checking.

That was found as needed as we found a big number of places where we do the same checks and an extra benefit is that
the user will no longer have to import Version from packaging in order to make the version comparable.

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -77,10 +77,7 @@ class Runtime:
             self.cache_dir = get_cache_dir(self.project_dir)
         self.config = AnsibleConfig()
 
-        if (
-            min_required_version is not None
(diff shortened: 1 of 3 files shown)


src/ansible_compat/runtime.py
```diff
@@ -156,6 +153,19 @@ def version(self) -> packaging.version.Version:
         msg = "Unable to find a working copy of ansible executable."
         raise MissingAnsibleError(msg, proc=proc)
 
+    def version_in_rage(
```
tadeboro, line 156: Maybe we should use the calmer version of `version_in_range` here? ;)

tadeboro (changes requested):

PR #67, opened 2021-09-02
Add install_collection_from_disk method

ssbarnea:
As ansible-galaxy can install collections from disk only with 2.10+, we add a new method that can be used with all supported versions of Ansible.

This also enables install_collection to have a force argument.

Related: https://github.com/ansible-community/molecule/pull/3225 https://github.com/ansible-community/ansible-compat/issues/65


src/ansible_compat/runtime.py
```diff
@@ -199,6 +203,31 @@ def install_collection(
             _logger.error(msg)
             raise InvalidPrerequisiteError(msg)
 
+    def install_collection_from_disk(
+        self, path: str, destination: Optional[Union[str, pathlib.Path]] = None
+    ) -> None:
+        """Build and install collection from a given disk path."""
+        if self.version_in_range(upper="2.11"):
```
webknjaz, line 210: Reverse this
```suggestion
        if not self.version_in_range(upper="2.11"):
            return self.install_collection(path, destination=destination)
```
and the rest of the code could be dedented.

src/ansible_compat/runtime.py
```diff
@@ -199,6 +203,31 @@ def install_collection(
             _logger.error(msg)
             raise InvalidPrerequisiteError(msg)
 
+    def install_collection_from_disk(
+        self, path: str, destination: Optional[Union[str, pathlib.Path]] = None
+    ) -> None:
+        """Build and install collection from a given disk path."""
```
webknjaz, line 209: Also, since you have destination passed down to all invocations the same way, consider defining a partial like this:
```suggestion
        """Build and install collection from a given disk path."""
        install_collection = functools.partial(self.install_collection, destination=destination)
```

src/ansible_compat/runtime.py
```diff
@@ -199,6 +203,31 @@ def install_collection(
...
+        if self.version_in_range(upper="2.11"):
+            # older versions of ansible able unable to install without building
+            with tempfile.TemporaryDirectory() as tmp_dir:
+                cmd = [
+                    "ansible-galaxy",
+                    "collection",
+                    "build",
+                    "--output-path",
+                    tmp_dir,
```
webknjaz, line 218: The value should reside next to the argument
```suggestion
                    "--output-path", tmp_dir,
```

tadeboro (dismissed):

src/ansible_compat/runtime.py
```diff
@@ -199,6 +203,31 @@ def install_collection(
...
+                    path,
+                ]
+                _logger.info("Running %s", " ".join(cmd))
+                run = self.exec(cmd, retry=True)
```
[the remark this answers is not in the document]
webknjaz: Agreed. I missed this but it makes total sense to me to do
```suggestion
                run = self.exec(cmd, retry=False)
```

src/ansible_compat/runtime.py
```diff
@@ -199,6 +203,35 @@ def install_collection(
             _logger.error(msg)
             raise InvalidPrerequisiteError(msg)
```
webknjaz, line 204: As for the failure with Ansible 2.9 in tests, maybe make the argument an explicit path? as in pass `./acme-goodies-1.0.0.tar.gz`?

ssbarnea: Sorted with full path.

src/ansible_compat/runtime.py
ssbarnea: I do not think is really needed, in would make the code harder to read, partials are not really the most popular python constructs.

src/ansible_compat/runtime.py
ssbarnea: Nope, black will reformat.

webknjaz: Then add ignores. It's ridiculous when a formatter forces keeping tightly coupled things in different places.

Issue #65, opened 2021-08-31
Failure to install local collection or role


greg-hellings:
When handling a role that's inside of a collection, ansible-compat is unable to determine that the collection exists. Therefore, it errors because a role name and namespace are not specified inside of `meta/main.yml`. [Here](https://github.com/oasis-roles/ansible_collection_system/pull/54/checks?check_run_id=3475530293) is an example CI run where this occurs.

ssbarnea:
`meta.main.yml` has nothing to do with that but the bug report is valid. At this moment the local collection is not installed so code trying to make use of the collection will fail. We will need to make several patches in order to implement this missing feature.

A collection presence will be determined by the presence of `galaxy.yml` in current directory. We are not yet sure if we want to make the local collection detection and installation implicit for `Runtime()` as as some consumers may not want this feature.

I am curious what @tadeboro and @webknjaz think about this. We could add a parameter (what name?) to either `Runtime()` constructor or the `Runtime.prepare_environment()` which would enable or disable the installation of local code (collection, maybe role for standalone roles). 

If we do this, what should be the default behavior install it if detected or not doing anything?

PR #69, opened 2021-09-04
Add ability to install local collection or role

ssbarnea:
- adds install_local parameter to prepare_environment(), defaulting  to False
- install local role or collection only when local_install is enabled

Fixes: #65

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -284,7 +284,10 @@ class Runtime:
                 raise AnsibleCommandError(run)
 
     def prepare_environment(
-        self, required_collections: Optional[Dict[str, str]] = None, retry: bool = False
+        self,
+        required_collections: Optional[Dict[str, str]] = None,
+        retry: bool = False,
+        install_local: bool = False,
     ) -> None:
         """Make dependencies available if needed."""
         if required_collections is None:
@@ -292,15 +295,30 @@ class Runtime:
 
         self.install_requirements("requirements.yml", retry=retry)
 
+        destination = f"{self.cache_dir}/collections" if self.cache_dir else None
         for name, min_version in required_collections.items():
             self.install_collection(
                 f"{name}:>={min_version}",
-                destination=f"{self.cache_dir}/collections" if self.cache_dir else None,
+                destination=destination,
             )
 
         self._prepare_ansible_paths()
-        # install role if current project looks like a standalone role
-        self._install_galaxy_role(self.project_dir, ignore_errors=True)
+
+        if not install_local:
+            return
+
+        if os.path.exists("galaxy.yml"):
+            # molecule scenario within a collection
+            self.install_collection_from_disk(".", destination=destination)
+        elif pathlib.Path().resolve().parent.name == 'roles' and os.path.exists(
+            "../../galaxy.yml"
+        ):
+            # molecule scenario located within roles/<role-name>/molecule inside
+            # a collection
+            self.install_collection_from_disk("../..", destination=destination)
+        else:
+            # no collection, try to recognize and install a standalone role
+            self._install_galaxy_role(self.project_dir, ignore_errors=True)
 
     def require_collection(  # noqa: C901
         self,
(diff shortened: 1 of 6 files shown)


src/ansible_compat/runtime.py
```diff
@@ -284,23 +284,36 @@ def install_requirements(self, requirement: str, retry: bool = False) -> None:
...
                 f"{name}:>={min_version}",
-                destination=f"{self.cache_dir}/collections" if self.cache_dir else None,
+                destination=destination,
             )
 
         self._prepare_ansible_paths()
-        # install role if current project looks like a standalone role
-        self._install_galaxy_role(self.project_dir, ignore_errors=True)
+
+        if install_local:
```
webknjaz, line 307:
```suggestion
        if not install_local:
            return
```

src/ansible_compat/runtime.py
```diff
@@ -380,18 +393,21 @@ def _prepare_ansible_paths(self) -> None:
...
+            else []
+        )
 
         for path_list, path, must_be_present in alterations_list:
-            if must_be_present and not os.path.exists(path):
-                continue
+            if not os.path.exists(path):
+                if must_be_present:
+                    continue
+                os.makedirs(path, exist_ok=True)
```
webknjaz, line 410: Why does this need `exist_ok`?

test/test_runtime.py
```diff
@@ -568,6 +564,7 @@ def test_runtime_env_ansible_library(monkeypatch: MonkeyPatch) -> None:
         ("9999.0", None, False),
         (None, "1.0", False),
     ),
+    ids=("1", "2", "3", "4", "5"),
```
webknjaz, line 567: These numbers are meaningless. Try something like
```suggestion
    ids=lambda lower, upper, expected: "{'>=' + lower + ', ' if lower else ''}{'<=' + upper + ', ' if upper else ''}: {'' if expected else 'non-'}matching".format(**locals()),
```

src/ansible_compat/runtime.py
ssbarnea: Destination folders may already exist, especially when using the library with isolated=False.

test/test_runtime.py
ssbarnea: too complicated, i only care to have a short test id. if it fails user will easily find which one failed.

src/ansible_compat/runtime.py
```diff
@@ -380,18 +395,21 @@ def _prepare_ansible_paths(self) -> None:
...
-            )
+        alterations_list.extend(
+            [
+                (roles_path, f"{self.cache_dir}/roles", False),
+                (library_paths, f"{self.cache_dir}/modules", False),
+                (collections_path, f"{self.cache_dir}/collections", False),
+            ]
+            if self.isolated
+            else []
+        )
```
tadeboro, line 406: Is there a reason for this non-functional change? As of now, this will only make `git blame` users jump through another hoop before they get to the proper result.

src/ansible_compat/runtime.py
```diff
@@ -284,23 +284,38 @@ def install_requirements(self, requirement: str, retry: bool = False) -> None:
...
-        self._install_galaxy_role(self.project_dir, ignore_errors=True)
+
+        if not install_local:
+            return
+
+        if os.path.exists("galaxy.yml"):
+            # molecule scenario within a collection
+            self.install_collection_from_disk(".", destination=destination)
+        elif os.path.exists("../../galaxy.yml"):
+            # molecule scenario located within a role part of a collection
```
tadeboro, line 314: This is not exactly true since the condition will trigger in all cases where the `galaxy.yml` is two levels up from the CWD and not only for roles.

src/ansible_compat/runtime.py
```diff
@@ -284,23 +284,38 @@ def install_requirements(self, requirement: str, retry: bool = False) -> None:
...
 
         self._prepare_ansible_paths()
-        # install role if current project looks like a standalone role
-        self._install_galaxy_role(self.project_dir, ignore_errors=True)
+
+        if not install_local:
+            return
+
+        if os.path.exists("galaxy.yml"):
+            # molecule scenario within a collection
```
tadeboro, line 311: This is not 100% correct since there is no rule that the `molecule` directory must live at the root of the collection.

src/ansible_compat/runtime.py
ssbarnea: Yes there is and it will make you laugh. Previous code was failing with C901 as function reached a cyclomatic complexity of 11. Rewriting that condition using the inline condition, tricket flake8 to believe complexity remained the same.

I personally prefer to add `# noqa: C901` to those methods/functions that need extra complexity but I am almost sure that @webknjaz would have requested a change if I went that approach.

src/ansible_compat/runtime.py
ssbarnea: We only ever supported at repository root and inside a role. I am keep to add support to add heuristics for finding any galaxy.yml at upper levels. It would add extra complexity and encourage people to invent new layouts, which will likely translate to new bugs.

Do you have a specific case that worked before and no longer works after we add this change?

tadeboro: I was just commenting on the comment here, not the functionality (the functionality is fine with me). But rereading it now, it does not bother me that much anymore, so I guess we can keep it as is.

tadeboro (commented):
The `test/collections/acme.goodies/acme-goodies-1.0.0.tar.gz` is part of this PR on purpose? Apart from that, PR look good to me.

tadeboro (approved):

PR #70, opened 2021-09-05
Minor test improvements

ssbarnea:
- change test method name to make it more explicit
- ensure caching does not affect testing of ansible_version
- remove empty galaxy.yml file from repository root (repo is not a broken collection)
- moved pip-compile hooks at the end to speedup pre-commit execution

### DIFF
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -94,4 +94,5 @@ def test_ansible_version_missing(monkeypatch: MonkeyPatch) -> None:
         MissingAnsibleError,
         match="Unable to find a working copy of ansible executable.",
     ):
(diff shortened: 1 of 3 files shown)


test/test_config.py
```diff
@@ -94,4 +94,5 @@ def test_ansible_version_missing(monkeypatch: MonkeyPatch) -> None:
         MissingAnsibleError,
         match="Unable to find a working copy of ansible executable.",
     ):
-        ansible_version()
+        # bypassing lru cache
```
webknjaz, line 97: Nit: try to put comments on the same line

PR #71, opened 2021-09-05
Avoid use direct use of os.environ

ssbarnea:
From now on, Runtime() will make a copy of os.environ and use that
instead, eventually making required modifications to it. This ensures
that modifications made by our runtime will not affect current
interpreter environment.

This sorts a bug observed during testing where modification to
os.environ started to cross-contaminate between tests.

Another benefit from this change is that it makes easier to adopt use
of containerized/remote ansible runtimes like execution engine. This
change should be considered major because it means that any consumer
of the library must use of `Runtime.exec()` to execute commands
using the modified environment. If they used subprocess.run or similar
they will need at least to pass Runtime.environ in order to avoid
execution errors as os.environ is no longer altered.

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -51,6 +51,7 @@ class Runtime:
         min_required_version: Optional[str] = None,
         require_module: bool = False,
         max_retries: int = 0,
+        environ: Optional[Dict[str, str]] = None,
     ) -> None:
         """Initialize Ansible runtime environment.
 
@@ -70,10 +71,13 @@ class Runtime:
                                 also perform Python imports from Ansible.
         :param max_retries: Number of times it should retry network operations.
                             Default is 0, no retries.
+        :param environ: Environment dictionary to use, if undefined
+                        ``os.environ`` will be copied and used.
         """
         self.project_dir = project_dir or os.getcwd()
         self.isolated = isolated
         self.max_retries = max_retries
+        self.environ = environ or os.environ.copy()
         if isolated:
             self.cache_dir = get_cache_dir(self.project_dir)
         self.config = AnsibleConfig()
@@ -125,6 +129,7 @@ class Runtime:
                 check=False,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
+                env=self.environ,
             )
             if result.returncode == 0:
                 break
@@ -391,11 +396,11 @@ class Runtime:
                 path_list.insert(0, path)
 
         if library_paths != self.config.DEFAULT_MODULE_PATH:
-            _update_env('ANSIBLE_LIBRARY', library_paths)
+            self._update_env('ANSIBLE_LIBRARY', library_paths)
         if collections_path != self.config.collections_paths:
-            _update_env(ansible_collections_path(), collections_path)
+            self._update_env(ansible_collections_path(), collections_path)
         if roles_path != self.config.default_roles_path:
-            _update_env('ANSIBLE_ROLES_PATH', roles_path)
+            self._update_env('ANSIBLE_ROLES_PATH', roles_path)
 
     def _install_galaxy_role(
         self, project_dir: str, role_name_check: int = 0, ignore_errors: bool = False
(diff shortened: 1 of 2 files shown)


ssbarnea:
@webknjaz I need to pick your brain on one thing. Both Runtime() and AnsibleConfig() to need access to our environ and both execute code using this environment.

I could try to pass Runtime.environ to AnsibleConfig in the constructor. Still this will not enable me to reuse Runtime.exec() inside config instance.

Another option would be to pass runtime instance when creating AnsibleConfig, but this will mean that those two classes will cross linked. I am not sure how big of a problem is that but at least it allows us to keep execution in one place, which would clearly be a big advantage towards enabling use of ansible-runner.

I already considered keeping environ only inside config, but that is not sorting the execution bit, only making a little bit harder to access via `runtime.config.environ` instead of the more direct `runtime.environ`.

WDYT?

src/ansible_compat/runtime.py
```diff
@@ -461,6 +466,22 @@ def _install_galaxy_role(
             link_path,
         )
 
+    def _update_env(self, varname: str, value: List[str], default: str = "") -> None:
+        """Update colon based environment variable if needed.
+
+        New values are added by inserting them to assure they take precedence
```
tadeboro, line 472:
```suggestion
        New values are prepended to make sure they take precedence
```

src/ansible_compat/runtime.py
```diff
@@ -461,6 +466,22 @@ def _install_galaxy_role(
...
+        over the existing ones.
+        """
+        if value:
```
tadeboro, line 475:
```suggestion
        if not value: return
```
We can then dedent the remainder of the function. But I would vote for dropping the condition since none of the calculations are especially expensive and I would assume that in the vast majority of use cases, the `value` parameter will not be an empty list.

src/ansible_compat/runtime.py
```diff
@@ -461,6 +466,22 @@ def _install_galaxy_role(
...
+            orig_value = self.environ.get(varname, default)
+            if orig_value:
+                # Prepend original or default variable content to custom content.
+                value = [*value, *orig_value.split(':')]
+            value_str = ":".join(value)
+            if value_str != self.environ.get(varname, ""):
+                self.environ[varname] = value_str
+                _logger.info("Added %s=%s", varname, value_str)
```
tadeboro, line 483: This message is misleading since adding variables is just one of the possibilities here. The other one is pretending information to the existing value.

src/ansible_compat/runtime.py
```diff
@@ -461,6 +466,22 @@ def _install_galaxy_role(
...
+    def _update_env(self, varname: str, value: List[str], default: str = "") -> None:
+        """Update colon based environment variable if needed.
+
+        New values are added by inserting them to assure they take precedence
+        over the existing ones.
+        """
+        if value:
+            orig_value = self.environ.get(varname, default)
+            if orig_value:
+                # Prepend original or default variable content to custom content.
```
tadeboro, line 478: This comment should be removed since it is wrong and the functions documentation already said what are we going to do.

tadeboro (changes requested):

src/ansible_compat/runtime.py
ssbarnea: I will use "Set" instead of "Added"

tadeboro:
> @webknjaz @tadeboro I need to pick your brain on one thing. Both Runtime() and AnsibleConfig() to need access to our environ and both execute code using this environment.
> 
> I could try to pass Runtime.environ to AnsibleConfig in the constructor. Still this will not enable me to reuse Runtime.exec() inside config instance.
> 
> Another option would be to pass runtime instance when creating AnsibleConfig, but this will mean that those two classes will cross linked. I am not sure how big of a problem is that but at least it allows us to keep execution in one place, which would clearly be a big advantage towards enabling use of ansible-runner.
> 
> I already considered keeping environ only inside config, but that is not sorting the execution bit, only making a little bit harder to access via `runtime.config.environ` instead of the more direct `runtime.environ`.
> 
> WDYT?

I would go in a slightly different direction and make `AnsibleConfig` a pure data container with no dynamic parts and move all process-execution parts to the `Runtime` object. Library consumers would then get their hands on the ansible config through the `config` property of the `Runtime` instance.

tadeboro (approved):

PR #72, opened 2021-09-06
Improve github actions

ssbarnea:
Instead of stopping execution of first failure, we run all
ansible tests before failing. This should give better errors as
you could spot if only a particular ansible version is affected.

Disable fail-fast for ansible testing so we can identify if
a particular python version is the only one affected.

Run unit testing only when linting is passing, saving resources as
linters is likely to fail often and is quite fast to run.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -93,7 +93,7 @@ jobs:
       ${{ matrix.tox_env }}@${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
-      # fail-fast: false
+      fail-fast: false
       # max-parallel: 5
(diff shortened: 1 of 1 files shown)


.github/workflows/tox.yml
```diff
@@ -92,8 +92,9 @@ jobs:
     name: >-
       ${{ matrix.tox_env }}@${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    needs: linters
```
webknjaz, line 95: This is not necessary, let everything run in parallel, this is the most responsive way + let's GHA platform schedule the jobs better.

webknjaz: Besides, linter failures does not mean that the functional changes cannot be tested/useful without perfect linter results.

PR #74, opened 2021-09-06
Make use of tox-gh-action

ssbarnea:


ssbarnea:
@webknjaz WDYT about that? Apparently grouping works but there are 4-5 lines that seem to not be part of the groups:
```
py36-devel recreate: /home/runner/work/ansible-compat/ansible-compat/.tox/py36-devel
py36-devel installdeps: ansible-core @ git+https://github.com/ansible/ansible.git
py36-devel develop-inst: /home/runner/work/ansible-compat/ansible-compat
py36-devel installed: ...
```

It is bit annoying but not a big deal.  Another question is around "recreate", i am not sure why I see recreate as the environment is already created at previous step and this action should not even run.

webknjaz:
> Another question is around "recreate", i am not sure why I see recreate as the environment is already created at previous step and this action should not even run.

This often depends on a combination of different args you set up when running tox. Sometimes it's a bug. Try checking the old tox runs to see if it's there too.

ssbarnea:
@webknjaz Take a look at https://github.com/ansible-community/ansible-compat/pull/74/checks?check_run_id=3526660101 

I think that is a proof that tox-gh-action fails on making the UI easier to use. The errored group is collapsed, not even visible from current screen and requires considerable scrolling in order to find an expand. Shortly that makes the entire plugin pretty much useless. I will raise it as bug, with the hope that some changes can be made to improve the output.

Blocked by https://github.com/ymyzk/tox-gh-actions/issues/80 and https://github.com/ymyzk/tox-gh-actions/issues/57

ssbarnea:
I will drop this endeavor as I do not like author action regarding adoption and I know that tox maintainer is already working on an action that will be released alongside tox v4.

It is better for us to wait till tox 4v is out.

Issue #40, opened 2021-07-27
Ensure exec() can be used in live move (tee?)


ssbarnea:
Once requirement for using exec() inside molecule is to be able to get real-time output from the executed command. That is because most molecule commands are running playbooks and the user wants to see which tasks are executed in real time.

For linter we have the other use-case where we do not want to show the output but we want to analyze the output (that is already covered with current implementation).

Molecule makes use of https://github.com/pycontribs/subprocess-tee/blob/main/src/subprocess_tee/__init__.py in order to do that but we could consider doing that ourselves and benefit from reducing the number of dependencies we have.

We can defer this feature for another release as for the moment molecule could continue to use subprocess-tee instead of using `Runtime.exec()` but longer term, we will need a solution, as it would prevent molecule from using execution environments if we do not to it.

ssbarnea:
@webknjaz WDYT about this?

webknjaz:
There should probably be a separate generator/method emitting the new lines.

PR #75, opened 2021-09-06
Add ability to execute commands in tee mode

ssbarnea:
This is a required feature for use within molecule as the executed
commands may run for a long period of time and users want to see
live progress.

Fixes: #40

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -8,9 +8,10 @@ import re
 import shutil
 import subprocess
 import tempfile
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 
 import packaging
+import subprocess_tee
 
 from ansible_compat.config import (
     AnsibleConfig,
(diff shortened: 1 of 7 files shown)


src/ansible_compat/runtime.py
```diff
@@ -116,14 +117,19 @@ def clean(self) -> None:
...
     def exec(
-        self, args: Union[str, List[str]], retry: bool = False
+        self, args: Union[str, List[str]], retry: bool = False, tee: bool = False
     ) -> CompletedProcess:
         """Execute a command inside an Ansible environment.
 
         :param retry: Retry network operations on failures.
         """
+        if tee:
+            runner: Callable[..., CompletedProcess] = subprocess_tee.run
```
webknjaz, line 127: Don't call it runner. By invoking it, it doesn't *just* create an object but also invokes commands. It must be a verb.

ssbarnea: `run_func` ?

src/ansible_compat/runtime.py
```diff
@@ -116,14 +117,19 @@ def clean(self) -> None:
             shutil.rmtree(self.cache_dir, ignore_errors=True)
 
     def exec(
-        self, args: Union[str, List[str]], retry: bool = False
+        self, args: Union[str, List[str]], retry: bool = False, tee: bool = False
```
webknjaz, line 120: I'm not sure how I feel about seeing a million args added to this method. It seems to be exhaustive.

ssbarnea: I doubt we will too many args here. In fact we already have less than original run, as the universal_newlines are implicit.

PR #77, opened 2021-09-13
Add env and cwd args to Runtime.exec()

ssbarnea:


test/test_runtime.py
```diff
@@ -619,9 +619,31 @@ def test_install_collection_from_disk_fail() -> None:
...
+def test_runtime_exec_cwd(runtime: Runtime) -> None:
+    """Check if passing cwd works as expected."""
+    cwd = "/"
+    result1 = runtime.exec(["pwd"], cwd=cwd)
+    result2 = runtime.exec(["pwd"])
+    assert result1.stdout.rstrip() == cwd
+    assert result1.stdout != result2.stdout
+
+
+def test_runtime_exec_env(runtime: Runtime) -> None:
```
webknjaz, line 639: Could you parameterize this?

ssbarnea: The thing is that I do not want to parametrize it, by doing it I only make things worse:

- harder to read the test logic
- extra complexity for enforcing parametrize on something that is not only about passing different arguments
- considerably more code in the end
- no speed benefits, in fact it would make it slower (even when parallelized)

Probably you seen that I use parametrize but that is not one of those places where I see it as adding clear benefits.

I will mark this as resolved.

cidrblock:
Code changes and tests look sane.  Safe to assume this is related to some other work, issue in another package?  Can the related PR or issue be linked?

cidrblock (approved):

PR #79, opened 2021-10-05
Add support for py310

ssbarnea:
Related: https://github.com/ansible-community/devtools/issues/6

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -107,6 +107,7 @@ jobs:
         - 3.7
         - 3.8
         - 3.9
+        - "3.10"
         # NOTE: Installing ansible under 3.10-dev is currently not
         # NOTE: possible because compiling cffi explodes.
         os:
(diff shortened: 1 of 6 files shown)


webknjaz (approved):

.github/workflows/tox.yml
```diff
@@ -117,22 +118,25 @@ jobs:
         include:
         - tox_env: py36
           os: ubuntu-20.04
-          python-version: 3.6
+          python-version: "3.6"
```
webknjaz, line 121: Please don't add extra quotes where they are unnecessary.

webknjaz (changes requested):

.github/workflows/tox.yml
ssbarnea: In this case I think that adding them is highly recommended because I already made the mistake of writing `3.10` which is converted to `3.1`. The correct value for this field is a string, not a float, so using quotes would be highly recommended.

This avoids future mistakes when someone would want to bump the version and they endup writing a float with zero at the end. Being consistent is better here.

PS. Obviously that I would not advocate for adding quotes to things that are obvious strings.

webknjaz:
> It would be nice to include a short notice in the commit message that we bumped the subprocess-tee dependency version because older ones have troubles with Python 3.10, but apart from that, things look OK to me.

Agreed, unrelated and unmentioned changes in patches are hard to pinpoint later.

PR #80, opened 2021-10-05
chore: update sphinx

ssbarnea:
- preparing for adding support for py3.10, as we will need sphinx 4.2.0 or newer.
- fix broken docs jobs due to incorrect URL

~Related: https://github.com/pypa/packaging.python.org/issues/995~
Related: https://github.com/ansible-community/devtools/issues/6

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -70,7 +70,7 @@ htmlhelp_basename = 'ansible-compat'
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
-    'packaging': ('https://packaging.python.org/en/latest/', None),
+    'packaging': ('https://packaging.pypa.io/en/latest/', None),
 }
 
 
--- a/docs/requirements.in
+++ b/docs/requirements.in
@@ -1,4 +1,4 @@
 sphinx-autobuild>=0.7.1,<1.0
-sphinx>=4.0.0,<5.0
+sphinx>=4.2.0,<5.0
 sphinx_ansible_theme
 myst_parser
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -68,7 +68,7 @@ six==1.16.0
     # via livereload
 snowballstemmer==2.1.0
     # via sphinx
-sphinx==4.1.1
+sphinx==4.2.0
     # via
     #   -r docs/requirements.in
     #   myst-parser


docs/conf.py
```diff
@@ -70,9 +70,11 @@
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
-    'packaging': ('https://packaging.python.org/en/latest/', None),
+    'packaging': ('https://packaging.python.org/', None),
```
webknjaz, line 73: This name would be more precise.
```suggestion
    'PyPUG': ('https://packaging.python.org', None),
```

docs/conf.py
```diff
@@ -70,9 +70,11 @@
...
 }
 
+# https://github.com/pypa/packaging.python.org/issues/995
```
webknjaz, line 76: This comment seems misplaced. Are you confusing the documentation site dedicated to the Python Packaging Guides vs the library called `packaging`?
```suggestion
```

docs/conf.py
```diff
@@ -70,9 +70,11 @@
...
+nitpick_ignore = [('py:class', 'packaging.version.Version')]
```
webknjaz, line 77:
```suggestion
nitpick_ignore = [
    ('py:class', 'packaging.version.Version'),  # autodoc autolinks type annotations, this one is not added
]
```

docs/conf.py
webknjaz: Looking closer, I think the correct fix would be to apply
```suggestion
    'packaging': ('https://packaging.pypa.io/en/latest', None),
```
(and maybe drop the ignore)

docs/conf.py
```diff
@@ -70,9 +70,13 @@
...
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
-    'packaging': ('https://packaging.python.org/en/latest/', None),
+    'packaging': ('https://packaging.python.org/', None),
 }
 
+# https://github.com/pypa/packaging.python.org/issues/995
+nitpick_ignore = [
+    ('py:class', 'packaging.version.Version'),  # autodoc autolinks type annotations, this one is not added
+]
```
webknjaz, line 79:
```suggestion
```

docs/conf.py
```diff
@@ -70,9 +70,13 @@
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
-    'packaging': ('https://packaging.python.org/en/latest/', None),
+    'packaging': ('https://packaging.python.org/', None),
 }
 
+# https://github.com/pypa/packaging.python.org/issues/995
+nitpick_ignore = [
+    ('py:class', 'packaging.version.Version'),  # autodoc autolinks type annotations, this one is not added
```
webknjaz, line 78: This exists as https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

webknjaz (approved):

docs/conf.py
```diff
@@ -70,7 +70,8 @@
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
-    'packaging': ('https://packaging.python.org/en/latest/', None),
+    'packaging-guide': ('https://packaging.python.org/', None),
```
webknjaz, line 73: is this really used anywhere?

webknjaz (commented):
I cannot approve of this title. It explains nothing.

webknjaz (changes requested):
urgh.. why is it so hard to withdraw an approval?

ssbarnea:
> I cannot approve of this title. It explains nothing.

Done, make it explicit.

webknjaz (approved):

PR #90, opened 2021-10-26
Add support for validating role names

ssbarnea:


src/ansible_compat/roles.py
```diff
@@ -0,0 +1,32 @@
+"""Wrapper class for Ansible Roles."""
+from keyword import iskeyword
+
+
+class Role:
+    """Utility class for validating roles."""
```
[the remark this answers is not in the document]
webknjaz: This is still a concern, I did not resolve it.

src/ansible_compat/roles.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __init__(self, name: str, check: bool = False):
+        """Instantiate an Ansible role."""
+        self.namespace = ""
+        if name.count(".") == 1:
+            self.namespace, self.name = name.split(".")
+        else:
+            self.name = name
+        if check and (name.count(".") > 1 or not self.validate()):
+            raise RuntimeError(f"Invalid role name: {self}")
```
tadeboro, line 19: I would just leave this code out and "force" users to call the `validate` method on the constructed instance. Calling validate here just forces me to handle bad roles in two places: once I construct the instance and once more when I call validate.

test/test_roles.py
```diff
@@ -0,0 +1,29 @@
...
+    """Check if Role.validate returns expected result."""
+    role = Role(role_name)
+    assert role.validate() is result
+
+    # asserts that exception is raised when check=True
+    if not result:
+        with pytest.raises(RuntimeError):
+            Role(role_name, check=True)
+    else:
+        Role(role_name, check=True)
```
tadeboro, line 29: This should live in a separate test (or be gone for good if we remove the `check` constructor parameter.

src/ansible_compat/roles.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def validate(self) -> bool:
+        """Return false if role is not properly named.
+
+        In the future other tests may be added to validate().
```
tadeboro, line 24: If we intend to expand the functionality of checks beyond name validation in the future, maybe it would be better to return a list of error messages instead of a boolean? This way consumers can supply users with more details when things will eventually fail to validate.

test/test_roles.py
ssbarnea: Here is a challenge: how to avoid duplicating all the 7 tests between two tests? I know that is good to separate tests but it is also good to avoid duplicating data and I do not fancy repeating all the 7 examples for two different tests.

src/ansible_compat/roles.py
ssbarnea: Yeah, i think it will be possible to remove it and keep the API simple. I will do it.

ssbarnea: I do not want to speculate on what may be needed in the future or not and make the api overly complex. If we will need messaging we can add another `errors()` method that returns False for valid roles or list of errors for invalid ones. 

validate() still expected to be true for valid roles, returning list of errors would still evaluate as true.

@webknjaz WTDY, should I replace `validate()` with `errors()` and reverse its meaning?

tadeboro: Validate has no predefined meaning (I would even consider it a bad name for a function that returns boolean). I usually go with `valid?` when programming in Ruby and `is_valid` in Python.

src/ansible_compat/roles.py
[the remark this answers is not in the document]
ssbarnea: I know that you prefer the functional implementation but the reason why I went for the extra complex  class based approach is that this would be really needed for validating roles as we would need extra context when creating them. Imagine that a role validity depends on the context, usually of an ee, or local ansible install. If it would be simple function that validates roles, it would easily end-up having an insane signature.

src/ansible_compat/roles.py
ssbarnea: I changed implementation based on received feedback, please review again.

src/ansible_compat/roles.py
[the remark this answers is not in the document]
webknjaz: It's just one context arg in the signature.

PR #92, opened 2021-11-02
Refactor use of pip-compile

ssbarnea:
Makes output of pip-compile consistent across machines by enforcing use of
the same python version when updating deps.


.pre-commit-config.yaml
```diff
@@ -11,6 +11,9 @@ ci:
   # https://github.com/pre-commit-ci/issues/issues/55
   - pip-compile
   - pip-compile-docs
+default_language_version:
+  # Needed in order to make pip-compile output predictable.
+  python: python3.9
```
webknjaz, line 16: Why not set this to the lowest supported version?

ssbarnea: In the past it was the best to use the newest python version, at least for tools like mypy and flake8. But as latest is 3.10 i can see that some developers may consider it harder to obtain than 3.6

That explains why is 3.9, as it used to be latest.

Now, I am not against switching to 3.6 for pip-compile on another change, especially if we find a reason for doing it.

PR #108, opened 2022-01-04
Fix release pipeline

ssbarnea:
Remove reusable release pipeline because we do not have access to environment secrets.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,43 @@ jobs:
   release:
     name: release ${{ github.event.ref }}
     needs: before-release
-    uses: ansible-community/devtools/.github/workflows/release.yml@main
-    with:
-      pypi: true
+    # unable to use environment with uses/with, basically cannot reuse release pipelines
+    environment: release
+    runs-on: ubuntu-20.04
+
+    env:
+      FORCE_COLOR: 1
+      PY_COLORS: 1
+      TOXENV: packaging
+      TOX_PARALLEL_NO_SPINNER: 1
+
+    steps:
+    - name: Switch to using Python 3.8 by default
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install tox
+      run: >-
+        python3 -m
+        pip install
+        --user
+        tox
+    - name: Check out src from Git
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # needed by setuptools-scm
+    - name: Build dists
+      run: python -m tox
+    - name: Publish to test.pypi.org
+      if: >-  # "create" workflows run separately from "push" & "pull_request"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -11,6 +11,43 @@ jobs:
...
+    - name: Publish to test.pypi.org
+      if: >-  # "create" workflows run separately from "push" & "pull_request"
+        github.event_name == 'release'
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.testpypi_password }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish to pypi.org
+      if: >-  # "create" workflows run separately from "push" & "pull_request"
+        github.event_name == 'release'
```
webknjaz, line 50: This doesn't correspond the comment

.github/workflows/release.yml
```diff
@@ -11,6 +11,43 @@ jobs:
...
+        python3 -m
+        pip install
+        --user
+        tox
+    - name: Check out src from Git
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0  # needed by setuptools-scm
+    - name: Build dists
+      run: python -m tox
```
webknjaz, line 40: Where's a separate step for env provisioning?

.github/workflows/release.yml
```diff
@@ -11,6 +11,43 @@ jobs:
   release:
     name: release ${{ github.event.ref }}
     needs: before-release
-    uses: ansible-community/devtools/.github/workflows/release.yml@main
-    with:
-      pypi: true
+    # unable to use environment with uses/with, basically cannot reuse release pipelines
```
webknjaz, line 14: do you have a demo of this?

.github/workflows/release.yml
ssbarnea: True, sorted.

.github/workflows/release.yml
ssbarnea: Try to add using the visual editor and you will see a nice error in the web editor. No need to run it, they added it to the live editor.

I am not sure if the called workflow can make use of its own environment secrets but if it does it would be big security issue as it would mean that callers would be able to use its secrets. Anyway we would not be able to re-use it as we have several accounts, one for each organization. Shortly, that seems to be a dead end.

Feel free to experiment with others ways of reusing the release pipelines, maybe we will find a solution. still, for the time being I am reverting to the old / known-to-work way. Lucky, I think that is the only repo where I attempted to reuse release pipeline.

.github/workflows/release.yml
ssbarnea: No need. I never used that separation for release. That only makes sense when you have lots of dependencies and/or multiple environments chained, not with two dependencies (wheel, twine).

PR #116, opened 2022-01-30
Remove black overrides

ssbarnea:


ssbarnea:
Merged without reviews because I am the only person that ever committed to this repo.

PR #120, opened 2022-02-25
Improve symlink recreation logic

ssbarnea:
This should avoid concurrency issues like seen on:
https://github.com/ansible-community/molecule/runs/5334560300?check_suite_focus=true


cidrblock (approved):

MAC-R0ch3:
@ssbarnea I can confirm that since 2.0.0, we are now getting the error you have linked. WIth version 1.0.0 of ansible-compat, things work fine. Here is the matrix that I've tested with Jenkins:

molecule 3.5.2 + ansible-compat 1.0.0: working
molecule 3.5.2 + ansible-compat 2.0.0: working
molecule 3.6.1 + ansible-compat 1.0.0: working
molecule 3.6.1 + ansible-compat 2.0.0: not working

The error message is similar to "FileExistsError: [Errno 17] File exists: '/home/runner/work/molecule/molecule/src/molecule/test/scenarios/driver/delegated' -> '/home/runner/.ansible/roles/molecule.delegated_test'"

MAC-R0ch3:
@ssbarnea For python 3.8 at least, we are having the issue that exists and is_symlink show different behaviors. 

Here is what the docu says:
If the path points to a symlink, [exists()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.exists) returns whether the symlink points to an existing file or directory.

ssbarnea:
I would gladly review pull requests. Keep in mind that ansible-compat 2.0 should be used with upcoming molecule v4. I would recommend adding a `ansible-compat<2.0` to molecule stable 3.x branch, so we avoid using it.

This week I am focusing on releasing ansible-lint v6, but once that goes out, I should have time to focus on molecule v4.

MAC-R0ch3:
Thanks @ssbarnea . I wasn't aware that ansible-compat 2.0 was targeted at molecule v4.x.
We'll then fix the ansible-compat version until we move to molecule v4.

PR #135, opened 2022-04-16
Install role in cache dir

### LINKED ISSUE #97: _install_galaxy_role fails on os.readlink(link_path) if `link_path = path / fqrn` is not a symlink
if this is run from a role directory where the role directory name is also name for the `fqrn`, you get this failure:

```
    self._install_galaxy_role(self.project_dir, ignore_errors=True)
  File "/home/user/.ansible/roles/limepepper.nginx/.venv/lib64/python3.9/site-packages/ansible_compat/runtime.py", line 412, in _install_galaxy_role
    if not exists or os.readlink(link_path) != str(target):
OSError: [Errno 22] Invalid argument: '/home/user/.ansible/roles/limepepper.nginx'
```

`os.readlink(link_path)` assumes that link_path is a symlink

https://github.com/ansible-community/ansible-compat/blob/acb665b713bd4fd9dcb96152f762d4224bad8825/src/ansible_compat/runtime.py#L494

which if you name your role directories with the `namespace`.`role_name` convention, it isn't
```
$ file /home/user/.ansible/roles/limepepper.nginx
/home/user/.ansible/roles/limepepper.nginx: directory
```

zhan9san:
Fix #97 

Would you kindly review it?

If both this PR and https://github.com/ansible-community/molecule/pull/3514 are merged, it would fix https://github.com/ansible-community/molecule/issues/3495 and https://github.com/ansible-community/molecule/issues/3464

As the `install_local` is hardcoded in https://github.com/ansible-community/molecule/pull/3409/files#r802489585

Before: the `ansible-compat` would install role into `~/.ansible/roles`.

After: The role will be installed in isolated `cache_dir`, e.g. `/Users/xxx/.cache/ansible-compat/496064/roles`.


ssbarnea (changes requested):
Can you please add at least on test for this change? Without adding any test we will likely break it in the future.

zhan9san:
Hi @ssbarnea 

Thanks for your review.

> Can you please add at least on test for this change? Without adding any test we will likely break it in the future.

It only changes the roles installation path, and the change is covered in the test below.
Do we need to add another kind of tests?

https://github.com/ansible/ansible-compat/pull/135/files#diff-5ed0c318d50bdbbff0c7d57b0d7663fe2aafd6b824390f204266d23de5ef6eb6R137-R142

If `isolated` is set to `True`, the installation path would be in the `cache_dir`. Otherwise, in the first `default_roles_path`.

Besides, a function doc is added to `_get_roles_path`.

ssbarnea:
Yes, I do want another test. Let me try explain what and why. Current patch is changing behavior for the library but without adding or altering current tests, this means that the newly desired behavior is not tested.

I want a test that would assert the desired behavior, one that would fail if we remove the code of your patch accidentally in the future.

I am not trying to make it harder for you, just to protect us from breaking it again in the future. With its current form, if I revert your code-changes, I am failing and tests. Tests are a way to prevent regressions.

If I understood the patch correct, you want to ensure that when using isolated the tested rule do not endup being installed in default ansible location but that is still found inside our own cache. When isolated is off, you want the other thing.

zhan9san:
Thanks. It's clear now.

I'll add it.

zhan9san:
Hi @ssbarnea 

I tried to add some, but finally the assertion would be always to verify whether the installation location(symlink) is correct(existed).

> Current patch is changing behavior for the library but without adding or altering current tests, this means that the newly desired behavior is not tested.

I am not sure you review the following snippet, https://github.com/ansible/ansible-compat/pull/135/files#diff-5ed0c318d50bdbbff0c7d57b0d7663fe2aafd6b824390f204266d23de5ef6eb6R137-R142

```
    if isolated:
        assert pathlib.Path(f"{runtime.cache_dir}/roles/{role_name}").is_symlink()
    else:
        assert pathlib.Path(
            f"{os.path.expanduser(runtime.config.default_roles_path[0])}/{role_name}"
        ).is_symlink()
```

> With its current form, if I revert your code-changes, I am failing and tests. Tests are a way to prevent regressions.

If the `path = self._get_roles_path()` is changed or removed, the test would fail.

Could you please give some clues more?

zhan9san:
> If I understood the patch correct, you want to ensure that when using isolated the tested rule do not endup being installed in default ansible location but that is still found inside our own cache. When isolated is off, you want the other thing.

This PR is to ensure the `isolated` is really `isolated`. Currently, no matter the `isolated` is set to `True` or `False`, the role would be installed in the default roles path instead of an isolated path.

With this PR, the role would be installed in an isolated path if `isolated` is set to `True` while it would be installed in default roles path if `isolated` is set to `False`.

Issue #137, opened 2022-05-09
Re-add/re-implement offline mode


MarkusTeufelberger:
This was removed (because it apparently already didn't work correctly before that) in https://github.com/ansible/ansible-compat/pull/37

For more context see https://github.com/ansible/ansible-lint/issues/2042 - functionality in `ansible-lint` is currently broken/bugged due to this library automatically exec'ing out to `ansible-galaxy` whenever a certain file exists in the root folder of a repository.

ssbarnea:
I agree that we need to bring back the offline mode in linter. The tricky question related to ansible-galaxy is should we assume that any install command requires online mode?

People could install archives from local disk, so it can become bit weird. Bug if we assume that `offline: true` skips any ansible-galaxy install call, it should not be hard to bring it back.

nre-ableton:
@ssbarnea Thanks for the quick fix!

ssbarnea:
That is not really fixed until we merge https://github.com/ansible/ansible-lint/pull/2115 but I need manual tests for that.

nre-ableton:
Well, it's at least a step in the right direction. :smile:

PR #138, opened 2022-05-09
Add offline option to prepare_environment

ssbarnea:
Fixes: #137

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -301,12 +301,14 @@ class Runtime:
         required_collections: Optional[Dict[str, str]] = None,
         retry: bool = False,
         install_local: bool = False,
+        offline: bool = False,
     ) -> None:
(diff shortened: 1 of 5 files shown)


ssbarnea:
@MarkusTeufelberger can you please check this one? Should it be enough?

MarkusTeufelberger:
At first glance it seems enough, I'll need to add it to `ansible-lint` first though (which needs to propagate/set the offline switch into `ansible-compat`, see https://github.com/ansible/ansible-lint/blob/eb5d96b30dbc4eaa7cca40dd90167ed986cc9021/src/ansiblelint/app.py#L251), earliest that I can do that is tomorrow.

I hope it still works with my use case of having a custom collection path where I vendor all collections, but that might expose a different issue (with isolation) or not be a problem at all. Thanks for the fast response! :-)

ssbarnea:
I will make a release and update the linter after.

PR #143, opened 2022-05-21
Add module to work with schemas

ssbarnea:
Based on original implementation made by @cirdblock this will allow
us to reuse schema tools across multiple project, especially as the
lower level python-jsonschema project does not produce identical
results between runs.


ganeshrn (approved):

priyamsahoo (approved):

cidrblock (approved):
Clever sorting method..... only because I learned about it the other day, if we ever need the others: https://docs.python.org/3/library/functools.html#functools.total_ordering

Issue #165, opened 2022-09-22
ERROR! Unexpected Exception, this is probably a bug: Cannot call rmtree on a symbolic link


ssbarnea:
Ansible galaxy collection install --force can fail if the target collection was symlinked. While this seems like a galaxy bug, we cannot rely on ansible-galaxy to fix it in all ansible versions, so we need a workaround.

```
 Installing 'cisco.nxos:0.0.0' to '/Users/ssbarnea/.cache/ansible-compat/15cb31/collections/ansible_collections/cisco/nxos'
E         to see the full traceback, use -vvv
E
E         ERROR! Unexpected Exception, this is probably a bug: Cannot call rmtree on a symbolic link
E
E         Command returned 250 code:
```

PR #166, opened 2022-09-22
Avoid galaxy install failure with already symlinked collections

ssbarnea:
Fixes: #165

### DIFF
--- a/src/ansible_compat/loaders.py
+++ b/src/ansible_compat/loaders.py
@@ -1,10 +1,29 @@
 """Utilities for loading various files."""
+from __future__ import annotations
+
+import os
 from typing import Any
 
 import yaml
+
+from ansible_compat.errors import InvalidPrerequisiteError
 
 
 def yaml_from_file(filepath: str) -> Any:
     """Return a loaded YAML file."""
     with open(filepath, encoding="utf-8") as content:
         return yaml.load(content, Loader=yaml.FullLoader)
+
+
+def colpath_from_path(filepath: str) -> str | None:
+    """Return a FQCN from a path."""
+    galaxy_file = f"{filepath}/galaxy.yml"
+    if os.path.exists(galaxy_file):
+        galaxy = yaml_from_file(galaxy_file)
+        for k in ("namespace", "name"):
+            if k not in galaxy:
+                raise InvalidPrerequisiteError(
+                    f"{galaxy_file} is missing the following mandatory field {k}"
+                )
+        return f"{galaxy['namespace']}/{galaxy['name']}"
+    return None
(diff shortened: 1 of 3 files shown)


src/ansible_compat/runtime.py
```diff
@@ -352,6 +352,20 @@ def prepare_environment(
             return
 
         if os.path.exists("galaxy.yml"):
+            if destination:
```
zhan9san, line 355: If the `destination` is None, the issue you want to work around also occur.

Do we need to consider the first default ansible_collection path(`~/.ansible/collections/`)?

src/ansible_compat/runtime.py
```diff
@@ -352,6 +352,20 @@ def prepare_environment(
...
+                fqcn = fqcn_from_path(os.getcwd())
+                if (
+                    fqcn
+                    and os.path.islink("{destination}/{fqcn}")
```
zhan9san, line 359: Do you miss `f`?

src/ansible_compat/loaders.py
```diff
@@ -1,10 +1,29 @@
...
     with open(filepath, encoding="utf-8") as content:
         return yaml.load(content, Loader=yaml.FullLoader)
+
+
+def fqcn_from_path(filepath: str) -> str | None:
+    """Return a FQCN from a path."""
+    galaxy_file = f"{filepath}/galaxy.yml"
+    if os.path.exists(galaxy_file):
+        galaxy = yaml_from_file(galaxy_file)
+        for k in ("namespace", "name"):
```
zhan9san, line 23: Do we need to consider the `author` keyword?

It is used in `_get_galaxy_role_ns`
 https://github.com/ansible/ansible-compat/blob/aac5ae2bd68bcf09d1dd7c37066072df30d98c2c/src/ansible_compat/runtime.py#L585-L587

src/ansible_compat/runtime.py
```diff
@@ -352,6 +352,20 @@ def prepare_environment(
...
+                    and os.path.realpath("{destination}/{fqcn}") != os.getcwd()
+                ):
+                    _logger.warning(
+                        "Collection is symlinked, but not pointing to %s directory. "
+                        "We will remove it.",
+                        os.getcwd(),
+                    )
+                    os.unlink("{destination}/{fqcn}")
```
zhan9san, line 367: Do we need to raise an error here instead of deleting it?

src/ansible_compat/runtime.py
ssbarnea: Sorted the none. I am not sure yet about the default. I preferred to enable this logic only for the case when the destination is specified, to limit the risks.

src/ansible_compat/runtime.py
ssbarnea: Removing a symlink does not destroy data. We log the action.

src/ansible_compat/loaders.py
ssbarnea: No, our code here is only for collections. Author fallback is used on roles.

PR #223, opened 2023-02-13
src/ansible_compat/runtime.py: Fix collection loading

apatard:
- Ensure the cache dir is added to the collection loader path, in order to allow linting a collection role against a module this collection is providing

- Add missing default collection path to the loader for ansible version prior 2.15.0.dev0, as done by ansible's _configure_collection_loader(). The only remaining difference is the missing handling of COLLECTIONS_SCAN_SYS_PATH.

See https://github.com/ansible/ansible-lint/issues/2969 for reference.

Signed-off-by: Arnaud Patard <apatard@hupstream.com>

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -144,9 +144,9 @@ class Runtime:
                 # noinspection PyProtectedMember
                 # pylint: disable=protected-access
                 col_path += self.config.collections_paths
-                col_path += [
-                    os.path.dirname(os.environ.get(ansible_collections_path(), "."))
-                ]
+                col_path += os.path.dirname(
+                    os.environ.get(ansible_collections_path(), ".")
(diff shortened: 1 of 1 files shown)


ssbarnea:
@apatard LGTM but can you please add a test? Just as safety measure to avoid regression.

apatard:
> @apatard LGTM but can you please add a test? Just as safety measure to avoid regression.

@ssbarnea I know I should have added two tests (1 for plugin in collection and 1 for the issue found in ansible-lint bug 2969) but I don't know enough the test suite to do that. If you can give me some clues about that, I can try adding them.

ssbarnea:
@apatard I will take it over from here as there are few other changes needed. Funny, I reached the same bug while debugging something else.

ssbarnea (approved):

Issue #258, opened 2023-05-02
4.0.1 : fails with project_dir.resolve() - 'str' object has no attribute 'resolve'


ianw:
One of our ansilbe-lint jobs has started failing with 4.0.1.  Last good run with 3.0.2 is [1], failure is [2].  Full traceback is

```
Traceback (most recent call last):
  File "/home/iwienand/programs/openstack-infra/openstack-zuul-jobs/.tox/linters/bin/ansible-lint", line 8, in <module>
    sys.exit(_run_cli_entrypoint())
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/iwienand/programs/openstack-infra/openstack-zuul-jobs/.tox/linters/lib/python3.11/site-packages/ansiblelint/__main__.py", line 266, in _run_cli_entrypoint
    sys.exit(main(sys.argv))
             ^^^^^^^^^^^^^^
  File "/home/iwienand/programs/openstack-infra/openstack-zuul-jobs/.tox/linters/lib/python3.11/site-packages/ansiblelint/__main__.py", line 157, in main
    initialize_options(argv[1:])
  File "/home/iwienand/programs/openstack-infra/openstack-zuul-jobs/.tox/linters/lib/python3.11/site-packages/ansiblelint/__main__.py", line 105, in initialize_options
    options.cache_dir = get_cache_dir(options.project_dir)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/iwienand/programs/openstack-infra/openstack-zuul-jobs/.tox/linters/lib/python3.11/site-packages/ansible_compat/prerun.py", line 13, in get_cache_dir
    basename = project_dir.resolve().name.encode(encoding="utf-8")
               ^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'resolve'
```

Note versions of `ansible-lint` (6.4.0) and `ansible` (2.12) didn't change between the two runs, just `ansible_compat`.

[1] https://storage.bhs.cloud.ovh.net/v1/AUTH_dcaab5e32b234d56b626f72581e3644c/zuul_opendev_logs_6e7/881890/1/check/openstack-zuul-jobs-linters/6e77077/job-output.txt
[...]

PR #259, opened 2023-05-02
Partial revert "Finish pathlib migration (#245)"

ianw:
This reverts commit 2c25aebfc7e1d313f30238a2bad8a7b68e952fc8 for the an incompatible change to src/ansible_compat_prerun.py, as the get_cache_dir() function is called externally with a string argument.

Closes: #258

### DIFF
--- a/src/ansible_compat/prerun.py
+++ b/src/ansible_compat/prerun.py
@@ -4,13 +4,13 @@ import os
 from pathlib import Path
 
 
-def get_cache_dir(project_dir: Path) -> Path:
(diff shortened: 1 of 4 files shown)


ianw:
It looks like reverting this is a bit harder than I thought, but I think the gist is right

ssbarnea:
The switch was on purpose, forcing consumers to do the switch. Just add a ceiling until they catch-up with the new code. Linter already fixed on main and release will be out soon. Molecule has a WIP PR to fix it too.

PR #264, opened 2023-05-11
Include original message in AnsibleWarning

ssbarnea:
Related: #3216


src/ansible_compat/runtime.py
```diff
@@ -122,7 +122,12 @@ def warning(
             formatted: bool = False,  # noqa: ARG001
         ) -> None:
             """Override ansible.utils.display.Display.warning to avoid printing warnings."""
-            warnings.warn(msg, category=AnsibleWarning, stacklevel=2)
+            warnings.warn(
+                message=msg,
+                category=AnsibleWarning,
+                stacklevel=2,
+                source={"msg": msg},
```
ssbarnea, line 129: The message argument is altered internally and we want the original string later.

PR #275, opened 2023-05-21
Improve requirements.yml detection

ssbarnea:
Gives explicit error message about invalid requirements.yml files, especially as an unsupported format was used during the 2020 split.

Closes: #274
Related: https://github.com/ansible-community/community-topics/issues/230


felixfontein:
Nit: the collection split was 2020, not 2021 :)

PR #278, opened 2023-05-23
Allow more control over verbosity when calling ansible-galaxy

ziegenberg:
This makes ansible-galaxy display important info when the logging level is increased. This helps with debugging ansible-lint.

When using `ansible-lint` with maximum verbosity level (`-vv`), it currently only outputs:

```
INFO     Running ansible-galaxy role install -vr roles/requirements.yml --roles-path /root/.cache/ansible-compat/d499d0/roles
INFO     Running ansible-galaxy collection install -v -r collections/requirements.yml
```

Outputting the `stdout` will help with debugging the installation of requirements.


ssbarnea (changes requested):
While I agree that we need to do something about it, we need to find a way to control verbosity from the consumer of compat library instead of harcoding the behavior.

ziegenberg:
Hi @ssbarnea, any thoughts on my proposed changes?

ziegenberg:
Hi @ssbarnea, any thoughts on my proposed changes?

I rebased my branch and solved the merge conflict.

PR #291, opened 2023-06-09
Add spec file

daaitudian:
Hi,
I added a spec file to help rpm build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ MANIFEST
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
 *.spec
+rpm/*.spec
 
 # Installer logs
 pip-log.txt
--- a/rpm/python-ansible-compat.spec.in
+++ b/rpm/python-ansible-compat.spec.in
@@ -1,0 +1,91 @@
+%global numcomm @numcomm@
+%global alphatag @alphatag@
+%global dirty @dirty@
+
+%global srcname ansible-compat
+%global pkgname python-ansible-compat
+
+%bcond_without tests
+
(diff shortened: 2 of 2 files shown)


ssbarnea:
@daaitudian Can you please give me some hints regarding how to build locally using this `spec.in` file? On linter repo I only use a spec file and build using `packit build locally`.

daaitudian:
Hi @ssbarnea :

Here is a brief explanation,

1. Fill in "Version", "Release" and "%Changelog" in the spec.in file according to the actual information

2. Pack the source code into the style required by "Source0"  then copy it to the Source folder. copy the rpm/python-ansible-compat.spec.in to the SPECS folder, and rename it to python-ansible-compat.spec

3. Use the “rpmbuild” command to perform the rpm build

Reference document: https://www.redhat.com/sysadmin/create-rpm-package

Do I need to write a detailed README ?

ssbarnea:
@daaitudian Please take a look at https://github.com/ansible/ansible-compat/pull/291#:~:text=2%20days%20ago-,Enable%20packit%20rpm%20packaging%20%23292,-Open which we will merge soon. It is based on the setup we already have for ansible-lint and we use packit to test rpm building, which is basically essential for knowing that the spec is valid.

Still, I know that my PR does not update the changelog, so I will be more than happy if you could make one or two follow-ups to improve it.

The reason why I did not took you change was because I was not able to test it, especially on proposed PRs. There is no need to add documentation if we can ensure that the testing happens on CI -- I would call this self-documented ;)

daaitudian:
@ssbarnea Thanks a lot for your help and modification. I will follow up on it later.

PR #294, opened 2023-06-27
Update the changelog

daaitudian:
Hi @ssbarnea,
In this PR, I added the changelog update, but the current problem is that I can't get the historical version information and write it to the changelog, so the changelog is not perfect. 
After I will find the more suitable method. If you have a better way, please let me know.


ssbarnea:
@daaitudian can you please help with getting packit to run? Content of changelog is less important than having rpm build.

ssbarnea:
Replaced by #315

PR #318, opened 2023-08-17
Add `sys.path` to collection paths

cidrblock:
Related: https://github.com/ansible/molecule/issues/4017

If not isolated and COLLECTION_SCAN_SYS_PATH is true, add the sys.paths to collection paths.

This is intended to mimic the functionality of the [collection loader](https://github.com/ansible/ansible/blob/9afaf2216b0bf38cfbd9a0c93409959f78d01820/lib/ansible/utils/collection_loader/_collection_finder.py#L254)

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -201,10 +201,9 @@ class Runtime:
         if "PYTHONWARNINGS" not in self.environ:  # pragma: no cover
             self.environ["PYTHONWARNINGS"] = "ignore:Blowfish has been deprecated"
 
-        self.config = AnsibleConfig()
-
         if isolated:
             self.cache_dir = get_cache_dir(self.project_dir)
+        self.config = AnsibleConfig()
 
         # Add the sys.path to the collection paths if not isolated
         self._add_sys_path_to_collection_paths()
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -9,6 +9,7 @@ import os
 import re
 import shutil
 import subprocess
+import sys
 import tempfile
 import warnings
 from collections import OrderedDict
@@ -200,9 +201,13 @@ class Runtime:
         if "PYTHONWARNINGS" not in self.environ:  # pragma: no cover
             self.environ["PYTHONWARNINGS"] = "ignore:Blowfish has been deprecated"
 
+        self.config = AnsibleConfig()
+
         if isolated:
             self.cache_dir = get_cache_dir(self.project_dir)
-        self.config = AnsibleConfig()
+
+        # Add the sys.path to the collection paths if not isolated
+        self._add_sys_path_to_collection_paths()
 
         if not self.version_in_range(lower=min_required_version):
             msg = f"Found incompatible version of ansible runtime {self.version}, instead of {min_required_version} or newer."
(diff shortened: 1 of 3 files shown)


test/test_runtime.py
```diff
@@ -444,6 +447,68 @@ def test_require_collection(runtime_tmp: Runtime) -> None:
...
+    monkeypatch.syspath_prepend(str(tmp_path))
+    runtime_tmp._add_sys_path_to_collection_paths()
+
+    try:
+        runtime_tmp.require_collection(
+            name="community.molecule",
+            version="0.1.0",
+            install=False,
+        )
+    except InvalidPrerequisiteError:
```
zhan9san, line 507: if there is no exception, the param.expected will not be used.
The parameterized is not precise

cidrblock: ha, good catch, TY

src/ansible_compat/runtime.py
```diff
@@ -231,6 +235,11 @@ def warning(
         # Monkey patch ansible warning in order to use warnings module.
         Display.warning = warning
 
+    def _add_sys_path_to_collection_paths(self) -> None:
+        """Add the sys.path to the collection paths."""
+        if not self.isolated and self.config.collections_scan_sys_path:
```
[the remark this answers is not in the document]
cidrblock: Since molecule calls with isolated False (https://github.com/ansible/molecule/blob/33e5d8bffb248c1013765e8347bb914d686d8038/src/molecule/app.py#L10) I wanted to keep the scope of the change to a minimum not know _exactly_ where isloated and why it might be used.

lint inits with "isolated=True" and I didn't want to impact it https://github.com/ansible/ansible-lint/blob/225366aa87636693e907204eeeba8cc1bd76e031/src/ansiblelint/app.py#L49

Maybe @ssbarnea has an opinion here?

ajinkyau (approved):
LGTM

src/ansible_compat/runtime.py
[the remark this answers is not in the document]
zhan9san: Here is my concern

For`sys.path`, as `ansible-galaxy` is invoked underneath, and only the path in `sys.path` contains a folder named `ansible_collections` is added in to `ansible-galaxy` search path.

To be exact, if `xxx/site-packages/ansible_collections` folder exists, we just need add `xxx/site-packages`. Otherwise, no `sys.path` should be added.

Besides `xxx/site-packages`, there are also some other path in sys.path, but I don't believe `ansible_collections` folder exists in the rest.

Please note `AnsibleCollectionConfig.collection_paths` in 
https://github.com/ansible/ansible/blob/863e2571db5d70b03478fd18efef15c4bde88c10/lib/ansible/cli/galaxy.py#L1647-L1649


https://github.com/ansible/ansible/blob/863e2571db5d70b03478fd18efef15c4bde88c10/lib/ansible/utils/collection_loader/_collection_finder.py#L254-L272

test/test_runtime.py
```diff
@@ -444,6 +447,70 @@ def test_require_collection(runtime_tmp: Runtime) -> None:
...
+        return "-".join(parts)
+
+
+@pytest.mark.parametrize(
+    ("param"),
+    (
+        ScanSysPath(isolated=True, scan=True, expected=False),
+        ScanSysPath(isolated=True, scan=False, expected=False),
+        ScanSysPath(isolated=False, scan=True, expected=True),
+        ScanSysPath(isolated=False, scan=False, expected=False),
```
zhan9san, line 472: Based on the result, if `scan` is False, no matter what value `isolated` is, the `expected` is always False.

If `scan` is `True`, I think the `expected` should be True. It's not related to isolated.

cidrblock: ```if not self.isolated and self.config.collections_scan_sys_path:```

this one confirms that if isolated and scan sys is true we still don't inject

```
        ScanSysPath(isolated=True, scan=False, expected=False),
```

cidrblock: I need @ssbarnea's input here, since lint uses isolated, I think he will better know the impact...  Maybe a sperate PR to remove the isolated logic and see what he says?

I do not have a good understanding of what isolated is intended to accomplish :/

PR #322, opened 2023-08-24
Add only those sys.paths which contain an ansible_collections directory path

ajinkyau:
Fixes: #323

Depends on #325

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -241,7 +241,7 @@ class Runtime:
             for path in sys.path:
                 if (
                     path not in self.config.collections_paths
-                    and Path(Path(path) / "ansible_collections").is_dir()
+                    and (Path(path) / "ansible_collections").is_dir()
                 ):
                     self.config.collections_paths.append(  # noqa: PERF401 pylint: disable=E1101
                         path,
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -8,6 +8,7 @@ import logging
 import os
 import re
 import shutil
+import site
 import subprocess
 import sys
 import tempfile
@@ -204,6 +205,11 @@ class Runtime:
         if isolated:
             self.cache_dir = get_cache_dir(self.project_dir)
         self.config = AnsibleConfig()
+
+        # Add the site packages path to the collection paths
+        self.config.collections_paths.extend(  # pylint: disable=E1101
+            site.getusersitepackages(),
+        )
 
         # Add the sys.path to the collection paths if not isolated
         self._add_sys_path_to_collection_paths()
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -243,7 +243,9 @@ class Runtime:
                     path not in self.config.collections_paths
                     and Path(Path(path) / "ansible_collections").is_dir()
                 ):
-                    self.config.collections_paths.append(path)  # pylint: disable=E1101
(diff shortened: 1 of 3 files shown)


ssbarnea (changes requested):
I would be far more confident about a test that asserts the fact that a collection that is installed under sitepackages is found after making this change instead of just checking the addition of the folder to the list.

What I do not know is how we can get a collection installed there in order to perform this test.

@cidrblock Any ideas?

zhan9san:
Plz kindly take a look at this comment
https://github.com/ansible/ansible-compat/pull/318#discussion_r1300846821

zhan9san:
Do we really need to add isolated in  condition here?

https://github.com/ansible/ansible-compat/pull/318#discussion_r1301889195

cidrblock (changes requested):
I think this PR fixes #323 but I don't see specifically how it addresses https://github.com/ansible/molecule/issues/4015. Can the title be modified please?

WRT to the change in tests, My preference is to test all permutations of the conditions evalutated in the code.

cidrblock:
> I would be far more confident about a test that asserts the fact that a collection that is installed under sitepackages is found after making this change instead of just checking the addition of the folder to the list.
> 
> What I do not know is how we can get a collection installed there in order to perform this test.
> 
> @cidrblock Any ideas?
> 
> The reason I am asking this is because I am not sure if the addition of the path really works and I find behavior test as adding more confidence than one that only does a unit-testing.

@ssbarnea You are 100% right. The test I added could be improved. The only reason to iterate sys.paths is because it exposes the site-packages directories where a full ansible might be found. I will PR a change to the test shortly.

cidrblock:
@zhan9san I have captured your concerns as new issues

#323 
#324

src/ansible_compat/runtime.py
```diff
@@ -238,7 +238,14 @@ def warning(
     def _add_sys_path_to_collection_paths(self) -> None:
         """Add the sys.path to the collection paths."""
         if not self.isolated and self.config.collections_scan_sys_path:
-            self.config.collections_paths.extend(sys.path)  # pylint: disable=E1101
+            for path in sys.path:
+                if (
+                    path not in self.config.collections_paths
+                    and Path(Path(path) / "ansible_collections").is_dir()
```
cidrblock, line 244: I think you may have an extra "Path" here :)

cidrblock (approved):
one comment

cidrblock:
> I would be far more confident about a test that asserts the fact that a collection that is installed under sitepackages is found after making this change instead of just checking the addition of the folder to the list.
> 
> What I do not know is how we can get a collection installed there in order to perform this test.
> 
> @cidrblock Any ideas?
> 
> The reason I am asking this is because I am not sure if the addition of the path really works and I find behavior test as adding more confidence than one that only does a unit-testing.

@ssbarnea I think #325 is a better test, LMKWYT

zhan9san:
> @cidrblock Because the issue states that the collection was installed in the site packages path which was not added to the collection_paths earlier.
Now since we're adding it, this will solve the issue.

I agree with @cidrblock that this PR only fixes the issue described in #323 .

And I think the issue described by @lod in https://github.com/ansible/molecule/issues/4015#issue-1852794321 is fixed by #318 

But the issue described by @isuftin in https://github.com/ansible/molecule/issues/4015#issuecomment-1680859724 is not related to this PR.

src/ansible_compat/runtime.py
ajinkyau: updated

Issue #324, opened 2023-08-25
Addition of site packages directories via sys.path limited to isolated


cidrblock:
As noted by @zhan9san the inclusion of sys.path entries is limited isolated=False.

The intent of this issue is to question the validity of this.

Related: https://github.com/ansible/ansible-compat/pull/318#discussion_r1301889195

PR #335, opened 2023-10-03
Remove isolated option while adding sys.path

shatakshiiii:
This PR validates the usage of `isolated` while adding sys.path to collections path.

Closes: https://github.com/ansible/ansible-compat/issues/324

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ concurrency = ["multiprocessing", "thread"]
 
 [tool.coverage.report]
 exclude_lines = ["pragma: no cover", "if TYPE_CHECKING:"]
-fail_under = 93
+fail_under = 91
 skip_covered = true
 show_missing = true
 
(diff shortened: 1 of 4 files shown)


ssbarnea (changes requested):
We do not want to lower the code coverage, removing of the test is a very bad idea and coverage was there to remind us not to make this mistake.

The correct solution for this is to modify the tests to check for the new behavior instead of removing them.

cidrblock:
Removing the conditional allows the number of tests to be dropped. As time permits, please add a test or 2 in a seperate PR to return coverage to it's previous value or higher.

cidrblock (approved):

shatakshiiii:
@ssbarnea  Also I realized that the changes in this PR is not responsible for lowering the code coverage, we could even see codecov failing in other open PRs. And dropping some unnecessary tests is suppose to happen as per these code changes.

PR #339, opened 2023-11-21
If project_dir is passed, locate requirements within that directory

audgirka:
When custom directory is passed using `project_dir`,
use it to locate the requirements file within that `project_dir` 
and install dependencies.
Related: [#3683](https://github.com/ansible/ansible-lint/issues/3863)


ajinkyau:
@ssbarnea How would `molecule` be impacted by this change?

ssbarnea (approved):

PR #345, opened 2023-12-29
Fix packit configuration

ssbarnea:


tox.ini
```diff
@@ -155,7 +155,7 @@ description = Use packit to build RPM (requires RPM based Linux distro)
 deps =
   packitos
 commands =
-  packit build in-mock
+  sh -c "packit build in-mock --root=fedora-40-$(arch)"
```
ssbarnea, line 158: Still looking for a trick to avoid hardcoded '40' here. Using 'latest' does not work.

PR #351, opened 2024-01-25
Fix Indentation

histamineblkr:
The `else` clause in `runtime.py` seemed to be missing one level of indentation. This PR fixes that.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,2 +1,3 @@
-# see https://github.com/ansible/devtools
-_extends: ansible/devtools
+---
+# see https://github.com/ansible/team-devtools
+_extends: ansible/team-devtools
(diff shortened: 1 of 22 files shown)


histamineblkr:
Ok I see a few issues with the pipeline.

1. Is I didn't use a key word like "bug", "enhancement", etc, which I can fix.
2. Is interesting: 
````
src/ansible_compat/runtime.py:724:9: RET503 Missing explicit `return` at the end of function able to return non-`None` value
src/ansible_compat/runtime.py:748:13: RET508 Unnecessary `else` after `break` statement
````
this `break` statement actually tells me it is unreachable, according to pylance, but I didn't dive into execution deep enough to feel confident in removing it. I attached a screenshot:
break-unreachable-pylance
3. I am uncertain why the various collections and things have broken in the `tox/py39...`

I am happy to fix anything on my side to get the pipeline fixed, but am unsure precisely your team would like to move forward.

PR #364, opened 2024-05-08
Add has_playbook method to Runtime

ssbarnea:
Add functionality for detecting if Ansible can find specific playbooks, caching the results. This is useful as they can reside inside collections.

Related: https://github.com/ansible/ansible-lint/pull/4141

### DIFF
--- a/test/test_runtime.py
+++ b/test/test_runtime.py
@@ -893,3 +893,16 @@ def test_get_galaxy_role_name_invalid() -> None:
         "role_name": False,  # <-- invalid data, should be string
     }
     assert _get_galaxy_role_name(galaxy_infos) == ""
+
+
+def test_runtime_has_playbook() -> None:
+    """Tests has_playbook method."""
+    runtime = Runtime(require_module=True)
+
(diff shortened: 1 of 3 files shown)


test/test_runtime.py
```diff
@@ -893,3 +893,16 @@ def test_get_galaxy_role_name_invalid() -> None:
         "role_name": False,  # <-- invalid data, should be string
     }
     assert _get_galaxy_role_name(galaxy_infos) == ""
+
+
+def test_runtime_has_playbook() -> None:
+    """Tests has_playbook method."""
+    runtime = Runtime(require_module=True)
+
+    assert not runtime.has_playbook("this-does-not-exit.yml")
```
Qalthos, line 902: Are these supposed to be `this-does-not-exist.yml` or are there some kind of halting problem shenanigans going on?

Qalthos (commented):
I like it otherwise

test/test_runtime.py
ssbarnea: Just a random name that I know not to exist on disk.

PR #374, opened 2024-05-10
Extend coverage

Qalthos:


src/ansible_compat/runtime.py
```diff
@@ -285,15 +285,19 @@ def load_collections(self) -> None:
...
+            raise RuntimeError(msg) from exc
         if not isinstance(data, dict):
             msg = f"Unexpected collection data, {data}"
             raise TypeError(msg)
         for path in data:
+            if not isinstance(data[path], dict):
+                msg = f"Unexpected collection data, {data[path]}"
+                raise TypeError(msg)
             for collection, collection_info in data[path].items():
-                if not isinstance(collection, str):
```
Qalthos, line 294: This is removed as I don't think it's possible to hit. Non-string keys won't get this far, as they won't be recognized as valid JSON and therefore hit the DecodeError above.

src/ansible_compat/runtime.py
```diff
@@ -772,18 +776,16 @@ def require_collection(
                             _logger.fatal(msg)
                             raise InvalidPrerequisiteError(msg)
                     return found_version, collpath.resolve()
-                break
-        else:
```
Qalthos, line 776: It doesn't look like this `break` can ever be reached, which means the else is also superfluous as anything that exits the loop does so by exhausting `paths`

ssbarnea (approved):

PR #443, opened 2025-01-07
Support ansible 2.18, drop 2.14/2.15 and py39

ssbarnea:
Related: https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -17,7 +17,7 @@ import warnings
 from collections import OrderedDict
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, no_type_check
+from typing import TYPE_CHECKING, Any, no_type_check
 
 import subprocess_tee
 from packaging.version import Version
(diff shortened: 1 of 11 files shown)


alisonlhart (approved):

src/ansible_compat/runtime.py
```diff
@@ -346,24 +347,15 @@ def _ensure_module_available(self) -> None:
...
-                )._install()  # pylint: disable=protected-access
+            # noinspection PyProtectedMember
+            # pylint: disable=protected-access
+            col_path += self.config.collections_paths
+            col_path += os.path.dirname(  # noqa: PTH120
+                os.environ.get(ansible_collections_path(), "."),
+            ).split(":")
+            _AnsibleCollectionFinder(  # noqa: SLF001
+                paths=col_path,
+            )._install()  # pylint: disable=protected-access
```
Qalthos, line 358: Not sure if I'm missing something, but this seems to be retaining the `< 2.15.0.dev0` path?

ssbarnea: Oops, read this after I merged it. Next time add -1. I am making a patch now. Second question would be.... why testing did not find any issues.

ssbarnea: Please review https://github.com/ansible/ansible-compat/pull/444

And, thanks for spotting it!!!

PR #582, opened 2026-04-19
fix: warn instead of fail on untested ansible-core/python combinations

eb4x:
ansible-core enforces its minimum Python through `requires-python` but declares no maximum, so pip will happily install an old ansible-core on a Python it was never tested with. #533 added a hard `RuntimeError` for the one case of Python 3.14 with ansible-core < 2.20.

This keeps that ceiling but changes two things:

1. **Warn, don't fail.** Older ansible-core releases mostly still work on newer Pythons. Running 2.16 through 2.19 on Python 3.14 here, fact gathering, templating and `copy` all succeed; what breaks is individual features that relied on stdlib modules Python has since removed, e.g. `password_hash` on 2.16 fails with `No module named 'crypt'`. That is worth a loud warning, but not a refusal to run `ansible-lint --version`.

2. **Derive the ceiling instead of special-casing 3.14.** ansible-core's stated policy is to support the three most recent Python releases at ship time, which with two core releases per year and one Python per year gives 2.16 → 3.12, 2.17 → 3.12, 2.18 → 3.13, 2.19 → 3.13, 2.20 → 3.14. The check now computes that, so it covers every combination without a table to maintain and won't need touching for 2.21 or 3.15.

The warning is a plain `warnings.warn` like the other checks in `prerun.py`, prefixed with `UNSUPPORTED:` so it stands out, and can be promoted to an error with `-W error` by anyone who wants the old behaviour.

Supersedes the earlier "invert the logic" approach in this PR, which was wrong: it asserted ansible-core 2.20 requires Python 3.14, and it dropped the ceiling entirely.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility checks for newer Python versions.
  - Environments running a Python version newer than the installed Ansible Core support range now receive a warning instead of an immediate failure.
  - The supported Python version is determined based on the installed Ansible Core version, providing more accurate compatibility guidance.


eb4x:
Hey @ssbarnea , can you please have a look at this. Trying to rectify an issue caused by #533. I'm not quite sure what do do here, mostly tempted to just remove it all, but you might better judge this.

The main point is this logic is flawed.

Python 3.14 doesn't depend on ansible-core, it doesn't care or even know about ansible-core.

Some versions of ansible-core might have a minimum runtime requirement of python. But even the tests here seem to run py312 with ansible-devel. So maybe python 3.14 isn't even the correct minimum requirement anymore?

shatakshiiii (dismissed):

shatakshiiii:
@eb4x code changes look good but related tests are failing, please address the tests failures so we can merge this PR. Moving this to draft until that's done.

 Thanks!

eb4x:
> @eb4x code changes look good but related tests are failing, please address the tests failures so we can merge this PR. Thanks!

That's why I had the question above "The tests here seem to run py312 with ansible-devel. So maybe python 3.14 isn't even the correct minimum requirement anymore?"

So should I lower the requirements until the tests pass, or remove it completely so the tests will pass?

ssbarnea (dismissed):
This is not correct because original test was preventing accidental use of unsupported old ansible-core versions with python 3.14, which can happen because ansible-core does not ceil python version but it does not support untested versions.

If you want you can extend the testing for both minimum and maximum versions but do not remove the the ceiling as that is on purpose.

eb4x:
> This is not correct because original test was preventing accidental use of unsupported old ansible-core versions with python 3.14, which can happen because ansible-core does not ceil python version but it does not support untested versions.

So you're saying `core_version >= Version("2.20.0dev0") and core_version < Version("2.20.1")` type situation?

ssbarnea:
@eb4x yes. If you really want you could make a matrix and check that python is inside supported range.

eb4x:
> @eb4x yes. If you really want you could make a matrix and check that python is inside supported range.

@ssbarnea thank you, and just so I get this straight, there's a tiny release window of ansible-core versions around 2.20 that had a hard requirement of python 3.14? like 2.20dev0 to 2.20dev4 ?

(I just want to make sure we're not saying something illogical like python3.14 can't run the python code in ansible-core-2.16)

shvenkat-rh:
@eb4x , do you have an update on this PR? it has been in draft for over 6 months

eb4x:
> @eb4x , do you have an update on this PR? it has been in draft for over 6 months

Yeah, sorry I wasn't getting any wiser on this. But thanks to "modern tooling", I think finally understand what @ssbarnea was getting at.

So I'll be changing the PR to ask for some leniency in how the check reacts: keep the ceiling, but go from a hard fail to a warning. Details in the updated PR.

shvenkat-rh (commented):
Good change overall — warn-not-fail is the right call here, and the dynamic formula is elegant. A few things worth discussing:

**`core_version` is now computed unconditionally at module import time**

Previously this was guarded by `if sys.version_info >= (3, 14)`, so the `importlib.metadata.version("ansible-core")` call only ran on Python 3.14+. It now runs on every import of `ansible_compat.prerun`. In practice this is fast and safe (if `ansible-core` isn't installed `ansible_compat` can't function anyway), but it's a behaviour change worth being aware of.

**`import platform` for a single format call**

`platform` is imported only to call `platform.python_version()` in the warning string. `sys` is already imported and `sys.version_info` is already used for the comparison, so you could drop the extra import:

```python
f"Python {sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
```

**Formula will need revisiting if Ansible changes release cadence**

`12 + (core_version.minor - 16) // 2` encodes a specific cadence (two core releases per year, one Python per year) and anchors to 2.16/3.12. The comment explains the reasoning well and the doc link helps, but the formula offers no defence against a silent wrong answer if that policy ever changes. A small table in the comment (like the one in the PR description) alongside the formula would make future audits much easier.

**No test coverage**

The `# pragma: no cover` is understandable given you'd need to fake both `sys.version_info` and the installed ansible-core version, but a test that mocks those two things and asserts a `UserWarning` is raised (and that no warning is raised for a supported combination) would cement the formula and catch regressions. Worth adding if not blocked by complexity.

**Everything else looks good.** The warning text is clear, `stacklevel=2` is consistent with the rest of `prerun.py`, and the formula itself is correct for all combinations from 2.16 onwards (verified against the Ansible support matrix).

shvenkat-rh (approved):

shatakshiiii (approved):

PR #583, opened 2026-05-03
fix: use absolute paths for roles and library in ANSIBLE_ROLES_PATH

Dotify71:
This PR fixes an issue where ansible-compat provides relative paths for roles and plugins in ANSIBLE_ROLES_PATH. These paths become invalid if the current working directory (CWD) is changed (which Molecule does when running scenarios). By using absolute paths derived from project_dir, we ensure they remain valid regardless of the CWD.

Related to ansible/molecule#4634 and ansible/molecule#3829


Dotify71:
This PR is a companion to the fix in Molecule: https://github.com/ansible/molecule/pull/4634. It ensures that role and plugin paths are resolved to absolute paths, preventing breakage when the working directory is changed (e.g., during Molecule scenario execution).

Dotify71:
Hello @ssbarnea! This PR is ready for review. It ensures absolute path resolution for roles and library discovery, preventing invalid paths during Molecule CWD changes. All 19 CI checks are passing, including new regression tests.

shatakshiiii (approved):
TY for the fix!

PR #606, opened 2026-07-16
fix: annotate mutable class defaults with ClassVar

rockygeekz:
## Summary
- Annotate all mutable class-level defaults (`list`, `dict`, `OrderedDict`) with `typing.ClassVar` in `AnsibleConfig` and `Runtime` classes
- Fixes 42 RUF012 violations that SonarCloud flags as code smells
- For `Runtime.collections`, removed the mutable class default since it is re-initialized per instance in `__init__`
- Left `AnsibleConfig.collections_paths` without `ClassVar` since it is reassigned through instances in `runtime.py`

## Test plan
- [x] All 127 existing tests pass locally
- [x] `ruff check src/` passes clean
- [x] `mypy` reports no new type errors
- [x] SonarCloud quality gate should turn green after merge

## Summary by CodeRabbit

* **Refactor**
  * Improved static type annotations for configuration defaults and runtime shared state.
  * Clarified class-level cache and default list/dict fields for better tooling and static analysis.
* **Bug Fixes**
  * Tightened version lookup error handling to only ignore missing package errors; other import issues now surface instead of being silently ignored.


Hrithik-Gavankar (approved):
Changes LGTM!

rockygeekz:
Using admin privilege to bypass the RTD checks because its not related to the PR changes and stuck

PR #608, opened 2026-07-17
fix: update noqa comments to new ruff syntax

rockygeekz:
Ruff 0.15.21 deprecated the old `# noqa: RULE` syntax in favor of `# ruff:ignore[RULE]` for inline and `# ruff:file-ignore[RULE]` for file-level suppressions. Update all 15 occurrences. This unblocks the pending pep621 dependency update PR #597 which is failing lint because of this.


rockygeekz:
Closing, the ruff syntax fix was already applied directly on #597 by ansibuddy. Will raise a new PR if main still needs the noqa migration after #597 merges.

PR #612, opened 2026-07-27
fix: skip collection modules in ANSIBLE_LIBRARY

### LINKED ISSUE #605: bug: Collection plugins/modules added to ANSIBLE_LIBRARY causes module shadowing in Molecule tests
## Issue Type

- Bug Report

## Summary

When testing an Ansible collection with Molecule, `ansible-compat` adds the collection's own `plugins/modules` directory to `ANSIBLE_LIBRARY`.

This exposes namespaced collection modules as legacy short-name modules. Collection modules can then unexpectedly shadow built-in Ansible modules used through `ansible.legacy`.

This regression is visible in `community.openwrt`, where `ansible.builtin.copy` internally invokes `ansible.legacy.stat`, but the loader resolves `stat` to:

```text
community.openwrt/plugins/modules/stat.py
```

instead of the built-in stat module.

The collection's `stat.py` is a documentation-only module stub intended to be used with its corresponding action plugin. When it is executed directly as a Python Ansible module, it produces no JSON output and the task fails with:

```text
Module result deserialization failed: No start of json char found
```

Example failing CI job:

https://github.com/ansible-collections/community.openwrt/actions/runs/29086551914/job/86341529977

## Regression source

The behaviour appears to result from the interaction of two changes.

### ansible-compat PR #583

PR:

https://github.com/ansible/ansible-compat/pull/583

Commit:

https://github.com/ansible/ansible-compat/commit/b258206e02244b9b5409877b8db9e130df5f5d1e

This changed the project module path added to `ANSIBLE_LIBRARY` from a relative path:

```python
(library_paths, "plugins/modules", True)
```

to an absolute path:

```python
(
    library_paths,
    str((self.project_dir / "plugins" / "modules").absolute()),
    True,
)
```

For a collection project, this becomes:

```text
.../ansible_collections/community/openwrt/plugins/modules
```

### Molecule PR #4634

PR:

https://github.com/ansible/molecule/pull/4634

Commit:

https://github.com/ansible/molecule/commit/74b2ec17bcc4a3c3e49e14eb98932cfd8b471ffe

This changed Molecule to preserve and merge the runtime environment generated by `ansible-compat`:

```python
env = (
    self.runtime.environ
    if env is None
    else util.merge_dicts(self.runtime.environ, env)
)
```

Together, these changes cause the collection's `plugins/modules` directory to be retained in `ANSIBLE_LIBRARY` when Molecule invokes `ansible-playbook`.

## Observed behaviour

The effective module search path in the failing job begins with the collection module directory:

```text
configured module search path =
[
  '/home/runner/work/community.openwrt/community.openwrt/ansible_collections/community/openwrt/plugins/modules',
  '/home/runner/.ansible/plugins/modules',
  '/usr/share/ansible/plugins/modules'
]
```

The failing task is:

```yaml
- name: Write inventory
  ansible.builtin.copy:
    dest: "{{ molecule_ephemeral_directory }}/inventory/hosts.yml"
    mode: "0644"
    content: "{{ {'all': {'hosts': _hosts}} | to_nice_yaml }}"
```

Before writing the file, the built-in copy action checks the destination using:

```python
module_name="ansible.legacy.stat"
```

Because the collection module directory is first in `ANSIBLE_LIBRARY`, the legacy lookup selects:

```text
.../community/openwrt/plugins/modules/stat.py
```

The CI log confirms this:

```text
Using module file
/home/runner/work/community.openwrt/community.openwrt/ansible_collections/community/openwrt/plugins/modules/stat.py
```

That module is then executed directly as `AnsiballZ_stat.py`, produces no JSON result, and fails with:

```text
Failed to get information on remote file:
Module result deserialization failed: No start of json char found
```

The inventory content itself is valid. The failure occurs during the destination stat operation before the inventory file is written.

## Expected behaviour

Collection plugin directories should remain within the collection namespace and be discovered through `ANSIBLE_COLLECTIONS_PATH`.

A collection directory such as:

```text
ansible_collections/community/openwrt/plugins/modules
```

should not automatically be treated as a legacy `ANSIBLE_LIBRARY` directory.

Collection modules should be resolved through their FQCN, for example:

```text
community.openwrt.stat
```

They should not automatically become available through the unqualified legacy name:

```text
stat
```

## Why this is problematic

Adding a collection's `plugins/modules` directory to `ANSIBLE_LIBRARY` removes the namespace boundary provided by collections.

Any collection module whose filename overlaps with a built-in or other legacy module can shadow that module during `ansible.legacy` lookup.

This issue is not limited to `stat` or `community.openwrt`. Other collections containing modules named `copy`, `file`, `command`, `shell`, `package`, or other common names could encounter similar behaviour.

Using an FQCN for the top-level task does not fully prevent the issue. In this example, the task uses `ansible.builtin.copy`, but the built-in copy action internally invokes `ansible.legacy.stat`.

## Suggested fix

`ansible-compat` should distinguish between traditional Ansible projects and Ansible collection projects when preparing runtime paths.

For collection projects, it should:

- Not add `<collection-root>/plugins/modules` to `ANSIBLE_LIBRARY`.
- Ensure the appropriate collection parent directory is available through `ANSIBLE_COLLECTIONS_PATH`.
- Preserve explicitly configured user and system `ANSIBLE_LIBRARY` entries.
- Continue adding project-local `plugins/modules` for non-collection projects where legacy module loading is expected.

Conceptually:

```python
if project_type != "collection":
    library_paths.insert(
        0,
        str((project_dir / "plugins" / "modules").absolute()),
    )
```

Molecule already detects collection mode, so it may need to communicate the project type to `ansible-compat`.

Alternatively, `ansible-compat` could detect a collection using project metadata such as `galaxy.yml`, while allowing callers to override that detection explicitly.

## Suggested regression test

Create a synthetic collection containing:

```text
ansible_collections/example/test/
├── galaxy.yml
└── plugins/
    └── modules/
        └── stat.py
```

Assert that:

```text
<collection-root>/plugins/modules
```

is not present in `ANSIBLE_LIBRARY`.

A Molecule integration test could then execute:

```yaml
- name: Test built-in copy operation
  ansible.builtin.copy:
    dest: /tmp/ansible-compat-module-resolution-test
    content: test
    mode: "0600"
```

The test should verify that the collection's `stat.py` is not selected during the built-in copy action's destination check.

## Versions observed

The failing environment installed:

```text
ansible-compat 26.6.0
molecule 26.6.0
ansible-core 2.19.11
```

Pinning Molecule below `26.6.0` avoids the regression in `community.openwrt`, but this is only a temporary workaround because the underlying collection path is still being treated as a legacy module library.

sudhirverma:
## Summary
- Skip injecting `project_dir/plugins/modules` into `ANSIBLE_LIBRARY` when the project has a `galaxy.yml` (collection projects).
- Keeps absolute library/roles path injection for non-collection projects (no regression of #583).
- Adds a regression test so collection module stubs (e.g. `stat.py`) cannot shadow `ansible.legacy` builtins.

Fixes #605

## Test plan
- [x] `uv run pytest test/test_runtime.py -k prepare_ansible_paths`
- [ ] Confirm Molecule collection CI (e.g. `community.openwrt`) no longer resolves `ansible.legacy.stat` to the collection stub after this release lands
- [ ] Spot-check a non-collection project still gets `plugins/modules` on `ANSIBLE_LIBRARY`

### DIFF
--- a/src/ansible_compat/runtime.py
+++ b/src/ansible_compat/runtime.py
@@ -970,13 +970,19 @@ class Runtime:
             raise RuntimeError(msg) from exc
 
         alterations_list: list[tuple[list[str], str, bool]] = [
-            (
-                library_paths,
-                str((self.project_dir / "plugins" / "modules").absolute()),
-                True,
-            ),
             (roles_path, str((self.project_dir / "roles").absolute()), True),
         ]
+        # Collection modules must stay namespaced; do not expose as legacy
+        # ANSIBLE_LIBRARY (see https://github.com/ansible/ansible-compat/issues/605).
+        if not (self.project_dir / GALAXY_YML).exists():
+            alterations_list.insert(
+                0,
+                (
+                    library_paths,
+                    str((self.project_dir / "plugins" / "modules").absolute()),
+                    True,
+                ),
+            )
 
         alterations_list.extend(
             (
--- a/test/test_runtime.py
+++ b/test/test_runtime.py
@@ -1133,6 +1133,42 @@ def test_prepare_ansible_paths_uses_absolute_library_path(
     )
 
 
+def test_prepare_ansible_paths_skips_library_for_collections(
+    tmp_path: pathlib.Path,
+) -> None:
+    """Collection plugins/modules must not be added to ANSIBLE_LIBRARY.
+
+    Regression test for https://github.com/ansible/ansible-compat/issues/605:
+    exposing a collection's modules as legacy short names can shadow builtins
+    (e.g. ansible.legacy.stat resolving to the collection's stat.py stub).
+
+    Args:
+        tmp_path: Pytest temporary directory fixture.
+    """
+    project_dir = tmp_path / "example_collection"
+    project_dir.mkdir()
+    (project_dir / "galaxy.yml").write_text(
+        "namespace: example\nname: test\nversion: 1.0.0\n",
+        encoding="utf-8",
+    )
+    modules_dir = project_dir / "plugins" / "modules"
+    modules_dir.mkdir(parents=True)
+    (modules_dir / "stat.py").write_text(
+        '"""Documentation-only module stub."""\n',
+        encoding="utf-8",
+    )
+
+    runtime = Runtime(project_dir=project_dir, isolated=False)
+    runtime._prepare_ansible_paths()
+
+    library_env = runtime.environ.get("ANSIBLE_LIBRARY", "")
+    project_library_path = str(modules_dir.absolute())
+    assert project_library_path not in library_env.split(":"), (
+        f"Collection modules path '{project_library_path}' must not appear in "
+        f"ANSIBLE_LIBRARY, got: {library_env}"
+    )
+
+
 def test_prepare_ansible_paths_survives_cwd_change(tmp_path: pathlib.Path) -> None:
     """Verify that resolved paths remain valid even after changing CWD.
 


shatakshiiii (approved):
LGTM!

